PR #336, opened 2026-04-16
improve(actions): drop fuel amount from Auto Fuel Toggle display (#335)

### LINKED ISSUE #335: improve(actions): drop fuel amount from Auto Fuel Toggle display
### What would you like?

Change the **Auto Fuel Toggle** key (the `toggle-autofuel` mode of the Fuel Service action) so it no longer renders the pit-service fuel amount (`PitSvFuel`) on the button. Make the visual treatment closer to **Fast Repair**: prominent title text and no live telemetry value drawn inside the key.

State is still indicated by the existing on/off/N/A status bar and border color — those stay. Only the `+X L` / `+X g` fuel text goes away.

### Use case

The original intent was to show the **iRacing Autofuel Lap Margin** on the button instead of the fuel amount. After checking, the lap-margin setting is **not exposed in iRacing telemetry** (no `margin` variable exists in `docs/reference/telemetry-vars.json`).

That leaves the fuel amount as the only dynamic value we could show, and it has limited usefulness on this particular key: the user can read `PitSvFuel` from the Fuel Fill toggle and other fuel actions. Removing it de-clutters the key and makes the on/off state the primary visual, matching how Fast Repair presents its state.

### Area

Action (existing)

### Alternatives considered

- **Show derived "laps of fuel remaining"** computed from `FuelLevel` ÷ fuel-used-last-lap. Rejected — it is not the same thing as iRacing's Autofuel Lap Margin setting and would be misleading on a key labelled "AUTO FUEL".
- **Leave the fuel amount in place.** Rejected — without lap margin there is no compelling reason to duplicate the fuel readout on this toggle.

### Additional context

**Files likely to change:**

- `packages/actions/src/actions/fuel-service.ts` — `generateFuelServiceSvg()` branch for `toggle-autofuel` (around lines 305–368). Stop calling `fuelFillGraphicContent()` for this mode; keep the status bar + border state logic. Consider rendering the static `toggle-autofuel.svg` graphic via the same path as Fast Repair (`assembleIcon()`) if that gives the cleanest result.
- `packages/actions/src/actions/fuel-service.test.ts` — update / add tests for the new `toggle-autofuel` icon output; the existing `toggle-fuel-fill` fuel-amount tests must remain unchanged.
- `packages/icons/fuel-service/toggle-autofuel.svg` — may need artwork content / updated title metadata so it reads well without a centered value on top of it. If artwork is added, also update `artworkBounds` in `<desc>` and regenerate previews via `node scripts/generate-icon-previews.mjs` (and `generate-artwork-bounds.mjs` if bounds change).
- `packages/stream-deck-plugin/src/pi/fuel-service.ejs` — no setting changes expected, but verify the `toggle-autofuel` mode still hides the fuel amount/unit inputs.
- Cache key in `FuelService.getStateKey()` (around line 759) — the `fuelAmount` / `displayUnits` fields are currently part of the `toggle-autofuel` state key and can be dropped, otherwise the key will regenerate the icon on every fuel tick for no visible change.
- Docs: `docs/plugins/core/actions/autofuel-toggle.md` and `packages/website/src/content/docs/docs/actions/pit-service/fuel-service.md` (Toggle Autofuel section) — update the icon-state description.

**Out of scope:** `toggle-fuel-fill` still shows `+X L` / `+X g`. Do not touch it.

**Reference implementation to mimic:** `pit-quick-actions.ts` Fast Repair mode + `packages/icons/pit-quick-actions/request-fast-repair.svg`.

niklam:
## Summary

- Remove the `+X L` / `+X g` fuel amount text from the `toggle-autofuel` mode so the key shows only the title and on/off status bar, matching Fast Repair's visual treatment
- Update the title to two-line "AUTO\nFUEL" at 18px font size with locked title fields
- Simplify the `buildStateKey()` to exclude `fuelAmount`/`displayUnits`, preventing unnecessary icon re-renders on every fuel tick

Closes #335

## Test plan

- [x] All 113 fuel-service tests pass (updated 3 tests to verify fuel amount is NOT shown)
- [x] 371 icon preview freshness tests pass
- [x] Full build passes (10/10 packages)
- [x] Manual verification: toggle-autofuel key shows "AUTO\nFUEL" title + status bar, no fuel amount text
- [x] Manual verification: toggle-fuel-fill still shows `+X L` / `+X g` as before

## Summary by CodeRabbit

* **Changes**
  * Autofuel toggle now displays status indicator only (ON/OFF), removing fuel amount display
  * Fuel-fill toggle retains fuel amount display functionality
* **Tests**
  * Updated test expectations for autofuel SVG rendering validation
* **Documentation**
  * Updated autofuel feature description to reflect status-only display behavior

### DIFF
--- a/packages/actions/src/actions/fuel-service.test.ts
+++ b/packages/actions/src/actions/fuel-service.test.ts
@@ -872,7 +872,7 @@ describe("FuelService", () => {
         expect(decoded).toContain("status-na");
       });
 
-      it("should show N/A status bar and -- text when autofuel system is not available", () => {
+      it("should show N/A status bar when autofuel system is not available", () => {
         const telemetryState: FuelServiceTelemetryState = {
           autofuelEnabled: false,
           autofuelActive: false,
@@ -885,11 +885,9 @@ describe("FuelService", () => {
         expect(decoded).toContain("status-na");
         expect(decoded).not.toContain("status-on");
         expect(decoded).not.toContain("status-off");
-        expect(decoded).toContain("--");
-        expect(decoded).not.toContain("+50 L");
       });
 
-      it("should include fuel amount graphic content", () => {
+      it("should not include fuel amount graphic content", () => {
         const telemetryState: FuelServiceTelemetryState = {
           autofuelActive: true,
           fuelAmount: 50.0,
@@ -898,15 +896,8 @@ describe("FuelService", () => {
         const result = generateFuelServiceSvg({ mode: "toggle-autofuel", amount: 1, unit: "l" }, telemetryState);
         const decoded = decodeURIComponent(result);
 
-        expect(decoded).toContain("+50 L");
-      });
-
-      it("should show -- when no fuel amount is available", () => {
-        const telemetryState: FuelServiceTelemetryState = { autofuelActive: true };
-        const result = generateFuelServiceSvg({ mode: "toggle-autofuel", amount: 1, unit: "l" }, telemetryState);
-        const decoded = decodeURIComponent(result);
-
-        expect(decoded).toContain("--");
+        expect(decoded).not.toContain("+50 L");
+        expect(decoded).not.toContain("--");
       });
 
       it("should include title content from metadata SVG", () => {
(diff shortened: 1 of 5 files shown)


niklam:
Good catch on the telemetry field reference — `PitSvFlags.FuelAutofill` was already wrong on master (pre-dates this PR). Fixed in 4e945b37 to reference `dpFuelAutoFillActive` which is what the code actually reads.

PR #338, opened 2026-04-17
fix(replay): cycle next/previous car on track during formation lap (#307)

### LINKED ISSUE #307: option for previous/next car based on track position instead of standings position
### What would you like?

currently the prev/next car switches based on standings position, in irace it, i use the buttons to switch between cars based on track position

### Use case

this is a convenience thing, it makes it much easier to review incidents 

### Area

Action (existing)

### Alternatives considered

_No response_

### Additional context

_No response_

niklam:
## Related Issue

Fixes #307

## What changed?

`findNearestCarOnTrack` (in `@iracedeck/iracing-sdk`) was filtering out every car whose `CarIdxLapCompleted < 0`. During a race's formation/pace lap no one has finished a lap yet, so the filter rejected every candidate and the replay `next-car` / `prev-car` commands became no-ops. Cycling by car number still worked because that path reads session info rather than telemetry lap counts — matching the reporter's and Niklas' observations.

- Dropped the lap-count filter (it duplicated the already-present `CarIdxLapDistPct >= 0` check for disconnected cars and wrongly rejected warmup-lap cars).
- Added an explicit `CarIdxTrackSurface !== NotInWorld` check as defense against disconnected / empty slots, since telemetry snapshots are only a single point in time and other states may set these fields inconsistently.
- Updated tests: split the former "skip inactive cars" case into explicit "no track position" and "NotInWorld" cases, and added a formation-lap regression test seeded from `telemetry-snapshot-20260417-081043` (all cars at `CarIdxLapCompleted = -1`, valid `CarIdxLapDistPct`).

## How to test

- Join a race and press the `Replay Control` action in `Next Car` / `Previous Car` mode while still on the warmup / formation lap (before any car has completed lap 1). The camera should now switch to the physically adjacent car.
- Verify mid-race behaviour is unchanged: cycle through cars ahead / behind on track during green-flag racing.
- Verify disconnected drivers are still skipped.
- `pnpm test` — 2909/2909 pass.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — fix is in shared `@iracedeck/iracing-sdk`, no per-plugin changes needed

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable car detection: disconnected vehicles are excluded and adjacent ahead/behind identification works during warm-up, pace, and formation laps.

* **Tests**
  * Expanded test coverage for disconnected/stale-position scenarios and for adjacent-car selection when lap-completion data is unavailable.

### DIFF
--- a/packages/actions/src/actions/replay-control.test.ts
+++ b/packages/actions/src/actions/replay-control.test.ts
@@ -1,4 +1,4 @@
-import { getAllCarNumbers, getCarNumberFromSessionInfo } from "@iracedeck/iracing-sdk";
+import { getAllCarNumbers, getCarNumberFromSessionInfo, TrkLoc } from "@iracedeck/iracing-sdk";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import {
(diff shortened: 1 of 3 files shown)


niklam:
@coderabbitai review

Issue #189, opened 2026-03-25
feat(actions): add Escape control type to Car Control action


niklam:
## What would you like?

Add a new **Escape** control type to the existing Car Control action. This sends the hardcoded ESC key (not configurable via global bindings) and includes an **Auto-hold** option.

### Behavior

- **Manual hold (auto-hold OFF)**: Press ESC on key down, release on key up — same pattern as Starter and Headlight Flash
- **Auto-hold ON**: Single tap presses ESC and automatically releases after 1.5 seconds, so users don't have to physically hold the button
- **Hardcoded ESC key**: Uses `getKeyboard()` directly — no global binding needed

### Settings

| Setting | Type | Default | Description |
|---------|------|---------|-------------|
| Control | Dropdown | Starter | Set to "Escape" for this feature |
| Auto-hold | Checkbox | Off | When enabled, a single tap simulates a 1.5s long press |

The Auto-hold checkbox should only be visible when the Escape control is selected.

## Use case

In iRacing, exiting the car requires holding the ESC key. Currently, users must physically hold the Stream Deck button for the full duration. An auto-hold option lets users tap once and have the system handle the hold duration automatically.

Additionally, this separates the ESC key functionality from the existing Enter/Exit/Tow control (which uses `Shift+R`), giving users more granular control over car exit behavior — similar to what iRaceIt offered.

## Alternatives considered

- Adding auto-hold to the existing Enter/Exit/Tow control — rejected because ESC and Shift+R serve different purposes in iRacing
- Making the key configurable via global bindings — rejected because ESC is always ESC in iRacing

### Implementation scope

Beyond the action code changes, the following must also be updated:

- **Icon**: New `packages/icons/car-control/escape.svg` (144x144, Mustache template)
- **PI templates**: Both `stream-deck-plugin` and `mirabox-plugin` car-control PI templates need the Escape option and conditional auto-hold checkbox
- **Tests**: `car-control.test.ts` — escape manual hold, auto-hold with timer, timer cleanup on disappear
- **Documentation**: `docs/reference/actions.json` and `iracedeck-actions` skill
- **Icon scripts**: Run `generate-icon-previews.mjs` and `generate-color-defaults.mjs` after icon creation

### Key technical notes

- Escape uses `getKeyboard().pressKeyCombination()` / `releaseKeyCombination()` directly (not the binding dispatcher pattern) since the key is hardcoded
[...]

PR #372, opened 2026-04-19
feat(actions): add Auto-hold options to Enter/Exit/Tow car control (#212)

### LINKED ISSUE #212: feat(actions): add auto-hold options to Enter/Exit/Tow car control
## What would you like?

Add three **Auto-hold** checkboxes to the Enter/Exit/Tow car control, one for each hold-based state: **Exit**, **Reset**, and **Tow**. Enter is instant (no hold needed), so it doesn't need one.

### Existing pattern

This functionality already exists in the **Escape** mode (#189), which has a single auto-hold checkbox. The Enter/Exit/Tow implementation is almost identical — the same auto-hold timer pattern can be reused — except it needs three independent checkboxes (one per hold state) instead of one.

### Behavior

- **Manual hold (auto-hold OFF, default)**: Press key on key down, release on key up — current behavior
- **Auto-hold ON**: Single tap presses the key and automatically releases after 1.5 seconds, so users don't have to physically hold the button

Each checkbox is independent, allowing users to mix manual and auto-hold per state. For example, a user might want auto-hold for Tow (which requires a long hold in race sessions) but prefer manual hold for Exit.

The active auto-hold setting is determined by the current telemetry state (`getEnterExitTowState()`):
- `enter-car` — no hold needed (instant action)
- `exit-car` — uses **Auto-hold Exit** setting
- `reset-to-pits` — uses **Auto-hold Reset** setting
- `tow` — uses **Auto-hold Tow** setting

### Settings

| Setting | Type | Default | Description |
|---------|------|---------|-------------|
| Auto-hold Exit | Checkbox | Off | Auto-hold when exiting the car (in pit stall) |
| Auto-hold Reset | Checkbox | Off | Auto-hold when resetting to pits (practice/qualifying) |
| Auto-hold Tow | Checkbox | Off | Auto-hold when requesting a tow (race session) |

All three checkboxes should only be visible when the **Enter/Exit/Tow** control is selected in the Car Control dropdown.

### Use case

In iRacing, exiting the car, resetting to pits, and requesting a tow all require holding Shift+R for a duration. Currently, users must physically hold the Stream Deck button. Auto-hold lets users tap once and have the system handle the hold duration automatically.

Different states may warrant different preferences — a user might always want quick tow access but prefer manual control when exiting.

## Alternatives considered

- A single auto-hold checkbox for all states — rejected because users may want different behavior per state (e.g., auto-hold for tow but manual for exit)
- Adding this to the Escape control from #189 — rejected because Enter/Exit/Tow uses Shift+R (global binding), not ESC

### Related

- #189 — Adds Escape control with auto-hold to Car Control (same timer pattern can be reused)

### Implementation scope

Beyond the action code changes, the following must also be updated:

- **Settings schema**: Add `autoHoldExit`, `autoHoldReset`, `autoHoldTow` boolean fields to `CarControlSettings`
- **Action logic**: Check telemetry state + corresponding auto-hold setting in `executeControl()` to decide between `holdBinding()` and a timed hold-then-release pattern
- **Timer management**: Track auto-hold timers per action context (`Map<string, ReturnType<typeof setTimeout>>`) with cleanup in `onWillDisappear` and `onKeyUp`/`onDialUp`
- **PI templates**: Both `stream-deck-plugin` and `mirabox-plugin` car-control PI templates need the three conditional checkboxes (visible only when `enter-exit-tow` is selected)
- **Tests**: `car-control.test.ts` — auto-hold per state, timer cleanup on disappear, manual hold still works when auto-hold is off
- **Documentation**: `docs/reference/actions.json` and `iracedeck-actions` skill

### Key technical notes

- Enter/Exit/Tow is already in the `HOLD_CONTROLS` set — the auto-hold logic wraps the existing hold pattern
- Uses the binding dispatcher (`holdBinding`/`releaseBinding`), not direct keyboard access, since the key is configurable via global settings
- Auto-hold timer should cancel any existing timer for the same action context before starting a new one
- `onKeyUp`/`onDialUp` should be a no-op when auto-hold is active (the timer handles release)

niklam:
## Related Issue

Fixes #212

## What changed?

Adds a per-state Auto-hold checkbox list to the Car Control action's Enter/Exit/Tow mode, mirroring the existing Escape auto-hold pattern from #189. Users opt in per telemetry-driven state:

- **Exit** (in pit stall)
- **Reset** (on track, non-race session)
- **Tow** (on track, race session)

Enter Car is instant and has no entry. The active auto-hold is resolved from live telemetry via `getEnterExitTowState()`, a single tap in an auto-hold state holds `Shift+R` for 1.5 s (shared `AUTO_HOLD_DURATION`), and a second tap cancels early. The shared `autoHoldTimers` map handles second-press cancellation, `onKeyUp`/`onDialUp` skip release while a timer is active, and `onWillDisappear` clears the timer before releasing the binding.

Implementation notes:
- Schema: `CarControlSettings.autoHoldStates` — `z.array` of `exit-car | reset-to-pits | tow` with a dedupe transform, mirroring the tire-service pattern.
- Routing: new `executeEnterExitTow()` mirrors `executeEscape()` but uses `holdBinding`/`releaseBinding` so SimHub routing still works.
- PI: single `sdpi-checkbox-list` (columns=1), shown only when Enter/Exit/Tow is selected in the Control dropdown.
- Tests, website docs, `docs/reference/actions.json`, and the `iracedeck-actions` skill updated.

## How to test

1. Build the repo (`pnpm build`) and reload the Stream Deck plugin (`streamdeck reload com.iracedeck.sd.core`).
2. Add a Car Control action, pick **Enter/Exit/Tow Car**. Confirm the **Auto-hold** row appears with three options (Exit / Reset / Tow) and disappears when you switch the control to anything else.
3. With all three unchecked: verify manual hold still works in each telemetry state (press-and-hold the button during each state — the key should be held while pressed and released on key up).
4. Check each option in isolation:
   - **Exit** in pit stall → single tap holds `Shift+R` for 1.5 s; a second tap during those 1.5 s cancels early.
   - **Reset** on track in a practice / qualifying session → same 1.5 s auto-hold behavior.
   - **Tow** on track in a race session → same 1.5 s auto-hold behavior.
5. In Enter Car state (out of car / replay / spectator), confirm auto-hold never engages even if all three options are checked.
6. Remove / re-add the action mid-hold to confirm `onWillDisappear` releases the binding cleanly.
7. Repeat on the Mirabox plugin (shared template, expected parity).

## Checklist

- [x] New code has unit tests

## Summary by CodeRabbit

* **New Features**
  * Added auto-hold functionality for Enter/Exit/Tow control, enabling single-tap activation with automatic 1.5-second release or re-press cancellation
  * Added configurable auto-hold per telemetry-driven state (Exit Car, Reset to Pits, Tow)

* **Documentation**
  * Updated Enter/Exit/Tow documentation with auto-hold behavior and configuration details

### DIFF
--- a/.claude/skills/iracedeck-actions/SKILL.md
+++ b/.claude/skills/iracedeck-actions/SKILL.md
@@ -66,7 +66,7 @@ Mode counts reflect the PI Mode/Setting dropdown choices documented in each acti
 | Audio Controls | 3 | push-to-talk (hold), voice-chat (with volume-up/down/mute action), master (with volume-up/down action) |
 | Black Box Selector | 3 | direct (with 11 Black Box options), next, previous |
 | Look Direction | 4 | look-left, look-right, look-up, look-down (all hold pattern) |
-| Car Control | 10 | pit-speed-limiter (telemetry-aware), push-to-pass (telemetry-aware), drs (telemetry-aware), headlight-flash (hold), tear-off-visor, ignition, starter (hold), enter-exit-tow (hold, telemetry-aware), escape (hardcoded ESC, auto-hold option), pause-sim |
+| Car Control | 10 | pit-speed-limiter (telemetry-aware), push-to-pass (telemetry-aware), drs (telemetry-aware), headlight-flash (hold), tear-off-visor, ignition, starter (hold), enter-exit-tow (hold, telemetry-aware, per-state auto-hold options for exit/reset/tow), escape (hardcoded ESC, auto-hold option), pause-sim |
 
 ### Cockpit & Interface
 
--- a/docs/reference/actions.json
+++ b/docs/reference/actions.json
@@ -121,7 +121,7 @@
             { "value": "tear-off-visor", "label": "Tear Off Visor" },
             { "value": "ignition", "label": "Ignition" },
             { "value": "starter", "label": "Starter", "controlType": "hold" },
-            { "value": "enter-exit-tow", "label": "Enter/Exit/Tow" },
+            { "value": "enter-exit-tow", "label": "Enter/Exit/Tow", "controlType": "hold", "telemetryAware": true },
             { "value": "escape", "label": "Escape", "controlType": "hold" },
             { "value": "pause-sim", "label": "Pause Sim" }
           ]
--- a/packages/iracing-actions/src/actions/car-control/car-control.ejs
+++ b/packages/iracing-actions/src/actions/car-control/car-control.ejs
@@ -25,6 +25,14 @@
 			<sdpi-checkbox setting="autoHold"></sdpi-checkbox>
 		</sdpi-item>
 
+		<sdpi-item id="auto-hold-states-item" label="Auto-hold" class="hidden">
+			<sdpi-checkbox-list setting="autoHoldStates" columns="1">
+				<option value="exit-car">Exit</option>
+				<option value="reset-to-pits">Reset</option>
+				<option value="tow">Tow</option>
+			</sdpi-checkbox-list>
+		</sdpi-item>
+
 		<%- include('title-overrides') %>
 
 		<script>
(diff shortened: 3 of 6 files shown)


niklam:
Thanks for the review.

The optional chaining in `getEnterExitTowAutoHold` is intentional defensive code to keep the test suite running. The real runtime path parses settings through the zod schema (where `.default([])` guarantees the array), so you're correct that in production `autoHoldStates` is always defined.

However, `car-control.test.ts` mocks `CommonSettings.extend` with a naive spread that bypasses zod entirely (the mock's `parse` just returns `{ ...data }`). Several tests intentionally pass partial settings like `{ control: "enter-exit-tow" }` to exercise defaults, and without the `?.` they would crash inside `.includes(...)`.

Options were:
1. Keep the `?.` (one extra character, zero runtime cost) — chosen.
2. Replace all tests to pass `autoHoldStates: []` explicitly.
3. Teach the mock to apply zod defaults (non-trivial for a single field).

Option 1 wins on signal-to-noise. Leaving as-is.

PR #423, opened 2026-04-23
improve(pit-crew): radar volume min 0 default 100 (#418)

niklam:
## Related Issue

Fixes #418.

## What changed?

Lower the radar-volume minimum from 5 to 0 so the slider alone can mute the radar without toggling the feature off. Supersedes the "clamps at 5 and 100" acceptance clause in #413.

- `packages/iracing-actions/src/actions/pit-crew/pit-crew.ts`: `VOLUME_MIN = 5` → `0`. `readRadarVolume()` already clamps via the constant.
- `packages/iracing-actions/src/actions/pit-crew/pit-crew.ejs`: `ird-range-input` `min="5"` → `min="0"`. Default stays at 100.
- `packages/deck-core/src/global-settings.ts`: `GlobalSettingsSchema.radarVolume: z.coerce.number().min(5).max(100)` → `.min(0)`. JSDoc updated.
- `pit-crew.test.ts`: "clamps at 5" → "clamps at 0"; added regression for step-down from 5 → 0.
- Docs sync: `docs/plugins/core/actions/pit-crew.md`, `docs/reference/actions.json`, `packages/website/.../pit-crew.md`, `.claude/skills/iracedeck-actions/SKILL.md`.

`VOLUME_STEP = 5` is deliberately preserved — stepping granularity is independent of the range.

## How to test

- [ ] PI: radar volume slider accepts 0–100. Default on a new button is 100.
- [ ] Radar Volume Down mode: stepping from 5 lands at 0 (no-op at 0).
- [ ] Live session: radarVolume=0 produces no ticks; 100 produces full-volume ticks.
- [ ] Existing configurations (typically 100) are unaffected.
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm -w test`, `pnpm format` all green (3168 tests).

## Checklist

- [x] Linked to an approved issue (#418, sub-issue of #375)
- [x] New code has unit tests (new step-to-0 regression)
- [x] Changes registered in all applicable plugins (Stream Deck rebuilds from shared EJS; Mirabox doesn't ship this action)
- [x] No unrelated changes included

Stacks on top of #422 (#417) which stacks on #421 (#416) which stacks on #420 (#415). Bases auto-retarget to `feature/pit-engineer` as the stack merges.

Sub-issue of #375.

## Summary by CodeRabbit

* **Bug Fixes**
  * Radar volume now accepts 0 as the minimum (0–100), allowing complete muting without disabling the radar; down-step behavior updated to reach 0.
[...]

PR #422, opened 2026-04-23
improve(pit-crew): replace raw HTML inputs with ird-* components (#417)

niklam:
## Related Issue

Fixes #417.

## What changed?

Pit Crew shipped two raw HTML inputs with hand-written JS wiring, violating `.claude/rules/stream-deck-actions.md` ("PI inputs must be `sdpi-*` or `ird-*`"). This PR replaces them with cohesive custom components and deletes the wiring.

New components in `packages/pi-components/src/components/`:

- **`ird-audio-test`** — styled button that writes `Date.now()` into a targetable `sdpi-textfield`, letting the action's `onDidReceiveSettings` detect the bump and fire `playRadarTest()`. Attributes: `target`, `label`.
- **`ird-audio-device-select`** — plugin-global device dropdown. Binds to the `audioOutputDevice` global and dynamically repopulates from `_audioDeviceList` (JSON) maintained by the plugin. Attributes: `setting`, `devices`, `default-label`.

Both follow existing `ird-range-input` / `ird-color-picker` / `ird-key-binding` patterns (once-guarded `connectedCallback`, shared style injection, `useGlobalSettings` hook, SSR-safe registration).

Updated:
- `pit-crew.ejs` accordion content: rewrites Radar Volume and Output Device rows to use the new components. The in-file `<script>` only keeps the Direction-visibility polling fallback.
- `stream-deck-actions.md`: documents both new components and adds a "never use raw `<button>/<select>/<input>/<textarea>`" clause to the Custom Components section.

## How to test

- [ ] Build succeeds; `pi-components.js` contains the two new class definitions; `ui/pit-crew.html` contains `<ird-audio-test target="test-radar-field" label="Test">` and `<ird-audio-device-select …>`.
- [ ] In Stream Deck, open a Pit Crew button's PI. The Test button plays the left → right → both sweep. Selecting an output device persists across PI reopens and is shared across every Pit Crew instance (global setting).
- [ ] `grep -n "<button\|<select\|<input\|<textarea" packages/iracing-actions/src/actions/pit-crew/pit-crew.ejs` returns zero hits.
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm -w test`, `pnpm format` all green. Test count grew from 3150 → 3167 (17 new cases covering DOM structure, late-arriving device list, malformed payload tolerance, settings persistence, and event bubbling).

## Checklist

- [x] Linked to an approved issue (#417, sub-issue of #375)
[...]

PR #421, opened 2026-04-23
chore(pit-crew): remove redundant mode-selector paragraph (#416)

niklam:
## Related Issue

Fixes #416.

## What changed?

Deletes the introductory `<p>` block above the Mode selector in `packages/iracing-actions/src/actions/pit-crew/pit-crew.ejs`. No other multi-mode action (force-feedback, cockpit-misc, camera-focus, session-info, black-box-selector) has an equivalent preamble — the Mode label is self-explanatory.

## How to test

- [ ] Open a Pit Crew button's PI. The section header flows directly into the Mode selector, matching the structure of other multi-mode actions.
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm -w test`, `pnpm format` all green.

## Checklist

- [x] Linked to an approved issue (#416, sub-issue of #375)
- [x] New code has unit tests (no new code; existing 3150 tests still pass)
- [x] Changes registered in all applicable plugins (shared EJS regenerates Stream Deck `ui/pit-crew.html`; Mirabox doesn't ship this action)
- [x] No unrelated changes included

Stacks on top of #420 (#415). Base will auto-retarget to `feature/pit-engineer` once #420 merges.

Sub-issue of #375.

## Summary by CodeRabbit

* **Documentation**
  * Removed explanatory guidance text from pit crew action settings regarding voice engineer and radar proximity alerts behavior.

Issue #418, opened 2026-04-23
improve(pit-crew): radar volume min 0 (was 5), default 100


niklam:
## Context

Sub-issue of #375. Lands on top of #413 (Pit Crew multi-mode refactor) on `feature/pit-engineer`. Supersedes the "clamps at 5 and 100" acceptance clause in #413.

The Radar Volume slider currently enforces a minimum of 5% (`VOLUME_MIN = 5`, `min="5"` on the `ird-range-input`). Sliding to the bottom still emits proximity ticks at 5% amplitude, which means the slider cannot mute radar output — the user has to press the Radar toggle to silence it. That mixes two concerns: the toggle enables/disables the feature entirely, while the slider controls loudness. Lowering the minimum to 0 lets the slider mute directly without touching the feature-enable state.

## Scope

- `packages/iracing-actions/src/actions/pit-crew/pit-crew.ts`
  - `const VOLUME_MIN = 5;` → `const VOLUME_MIN = 0;`
  - `readRadarVolume()` already uses `VOLUME_MIN` in its clamp — no additional edit needed.
- `packages/iracing-actions/src/actions/pit-crew/pit-crew.ejs`
  - `<ird-range-input setting="radarVolume" global min="5" max="100" step="1" default="100" …>`: `min="5"` → `min="0"`. `default="100"` stays.
- `packages/deck-core/src/global-settings.ts`
  - `GlobalSettingsSchema.radarVolume`: `z.coerce.number().min(5).max(100)` → `z.coerce.number().min(0).max(100)`.
- `pit-crew.test.ts`: update any assertion referencing `min=5` or the 5% clamp.
- `VOLUME_STEP = 5` stays — step granularity is still appropriate.

## Acceptance

- [ ] Radar Volume slider accepts 0–100 inclusive.
- [ ] Default on a new Pit Crew button is 100.
- [ ] Setting radarVolume to 0 silences `AudioBus.Alerts` completely (no radar ticks during a live session).
- [ ] Existing persisted configurations (typically 100) are unaffected.
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm format`, `pnpm -w test` all green.

## Commit message

`improve(pit-crew): radar volume min 0 default 100 (#<this issue>)`

Sub-issue of #375.

PR #424, opened 2026-04-23
fix(deck-core): update global-settings cache synchronously on local writes (#419)

### LINKED ISSUE #419: fix(pit-crew): radar toggle only disables, never re-enables; key stays green
## Context

Sub-issue of #375. Bug reproduced in a live iRacing session on `feature/pit-engineer`. Root cause sits in `@iracedeck/deck-core` global settings cache; Pit Crew is the observable surface because it's the only action that toggles a global flag inside `onKeyDown`.

## Symptoms

First press of a Pit Crew button in `radar` mode (while radar is enabled) silences the directional ticks correctly. **Every subsequent press also silences** — the toggle never flips back to enabled. The key image stays green (the "on" status bar colour) throughout.

## Repro log

Excerpt from a live session (times shortened, iRacing connected, radar ticking):

```text
10:13:15..10:13:20  (radar ticks firing every ~265 ms)
10:13:21  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 1 — correct (was true → false)
10:13:23  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 2 — should have flipped to true
10:14:40  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 3 — still false
10:14:41  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 4
10:14:43  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 5
10:14:44  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 6
10:15:07  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 7
10:15:09  PitCrew:        Radar disabled
          GlobalSettings: Partial update: {"radarEnabled":false}    # press 8

... ~4 minutes later, PI reopened ...

10:19:55  GlobalSettings: Settings received
          raw: "radarEnabled":false                                  # host finally echoes the 10:13:21 write
```

Every press computes `next = false` and writes `{"radarEnabled":false}`. The status bar on the key never flips to red because the icon generator reads the same stale value.

## Root cause

`packages/deck-core/src/global-settings.ts` only mutates the `currentSettings` cache inside the `onDidReceiveGlobalSettings` listener callback — never synchronously when `updateGlobalSettings()` is called locally. Trace:

1. Cache at mount: `{ radarEnabled: true }` (from host).
2. Press 1 → `isRadarEnabled()` reads cache → `true` → `toggleRadar()` computes `next = !true = false` → adapter write dispatched, but cache still reads `true`. Audio subsystem receives the new value via listener/side-effect path and goes silent.
3. Press 2 (100 ms later) → cache still `true` (no echo yet) → `next = !true = false` again → writes `false` again.
4. Steps repeat indefinitely until the host eventually broadcasts an `onDidReceiveGlobalSettings` echo. In Stream Deck the echo is not guaranteed synchronous — the user's log shows a ~6-minute gap before the cache was reconciled, at which point they had stopped pressing.

`generatePitCrewSvg()` → `modePresentation()` reads `isRadarEnabled()` for the status-bar colour, so the icon stays green for the same reason.

`toggleRaceEngineer()` has the symmetric bug but is invisible today because no audio/visual side-effect makes it observable.

## Scope

- **`packages/deck-core/src/global-settings.ts`** — fix `updateGlobalSettings(partial)` to merge `partial` into `currentSettings` and notify listeners **synchronously**, before (or in parallel with) handing the patch to the adapter. The later `onDidReceiveGlobalSettings` echo then parses + reconciles as a no-op (Zod should produce the same coerced value).
- **`packages/deck-core/src/global-settings.test.ts`** — add regression covering:
  1. `initGlobalSettings` with a mock adapter that *never* echoes back.
  2. `updateGlobalSettings({ radarEnabled: false })` → `getGlobalSettings().radarEnabled === false` immediately.
  3. `updateGlobalSettings({ radarEnabled: true })` → `getGlobalSettings().radarEnabled === true` immediately.
  4. Subscribed listeners fire on every local update (not only on host echoes).
- **`packages/iracing-actions/src/actions/pit-crew/pit-crew.test.ts`** — add a regression exercising press → press → press on the Radar mode: assert `radarEnabled` global alternates and the generated icon status bar alternates on each press, without an echo between presses.
- No change needed to `toggleRadar` / `toggleRaceEngineer` call sites — the root fix makes them correct.

## Acceptance

- [ ] Pressing a Radar Pit Crew button while radar is on silences the ticks AND flips the key image to red/off.
- [ ] Pressing again immediately brings the ticks back AND flips the key image to green/on.
- [ ] Alternation works indefinitely without waiting for an `onDidReceiveGlobalSettings` round-trip.
- [ ] `toggleRaceEngineer()` shows the same correct toggle behaviour (verified by unit test on the `raceEngineerEnabled` global).
- [ ] `updateGlobalSettings` unit test covers the synchronous cache update.
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm format`, `pnpm -w test` all green.

## Commit message

`fix(deck-core): update global-settings cache synchronously on local writes (#<this issue>)`

Sub-issue of #375.

niklam:
## Related Issue

Fixes #419.

## What changed?

`updateGlobalSettings()` used to hand the merged payload to the adapter and rely on the host's `onDidReceiveGlobalSettings` echo to refresh the in-memory cache. The Stream Deck host does **not** guarantee a prompt echo — in the session log that prompted this issue, the echo arrived **~6 minutes** after the write. Between the write and the echo, `getGlobalSettings()` returned stale state, so every read-modify-write toggle got stuck computing the same "next" value repeatedly.

This PR:

- Parses the merged partial via Zod and updates `currentSettings` + notifies listeners **synchronously** in `updateGlobalSettings`, before dispatching to the adapter.
- Extracts `applyParsedSettings(parsed)` as the single cache-mutation point; both the echo path and the local-write path funnel through it. No duplicate notifications; the later echo re-parses the same payload and reconciles as a no-op.
- Forwards the **parsed** (coerced) payload to `adapterRef.setGlobalSettings` so host and cache stay consistent.

Tests:
- **New** `packages/deck-core/src/global-settings.test.ts` — 8 cases covering single-write visibility, 3× alternation without echo, the Pit Crew read-flip-write shape, `raceEngineerEnabled` (same latent bug fixed), listener notification on local writes, parsed payload forwarding to the adapter, echo reconciliation after stale local writes, and uninitialised no-op guard.
- `pit-crew.test.ts` regression for the reported symptom: 3 consecutive Radar-mode presses alternate `radarEnabled` without a host echo (expected `true` → `false` → `true` → `false`).

## How to test

Live session:
- [ ] Connect to iRacing with radar ticks active.
- [ ] Press the Pit Crew Radar button once — ticks stop, key flips red.
- [ ] Press again immediately — ticks resume, key flips green.
- [ ] Alternate 4+ times without waiting for a PI round-trip; behaviour holds.
- [ ] Race Engineer toggle also alternates (status bar flips) across consecutive presses.

Automated:
- [ ] `pnpm -w build`, `pnpm -w lint`, `pnpm -w test`, `pnpm format` all green. Test count 3168 → 3177 (+9: eight new global-settings cases + one Pit Crew regression).

## Checklist

- [x] Linked to an approved issue (#419, sub-issue of #375)
- [x] New code has unit tests (new `global-settings.test.ts`, new Pit Crew regression)
- [x] Changes registered in all applicable plugins (fix is in `@iracedeck/deck-core`; both plugins consume the updated behaviour automatically)

Stacks on top of #423 (#418) → #422 (#417) → #421 (#416) → #420 (#415). Bases auto-retarget to `feature/pit-engineer` as the stack merges.

Sub-issue of #375.

## Summary by CodeRabbit

* **Bug Fixes**
  * Global settings now apply immediately in-memory and notify listeners synchronously on local changes, then reconcile with later host echoes to avoid stale or stuck states (fixes radar toggle flakiness).

* **Tests**
  * Added tests verifying immediate local-write behavior, listener notifications on every local change, host-echo reconciliation, and correct radar-mode toggle behavior.

### DIFF
--- a/packages/deck-core/src/global-settings.test.ts
+++ b/packages/deck-core/src/global-settings.test.ts
@@ -142,14 +142,11 @@ describe("global-settings cache (synchronous update on local writes)", () => {
 
   it("no-ops when the adapter is not initialized", () => {
     _resetGlobalSettings();
-    const logger = createMockLogger();
     // Not calling initGlobalSettings — adapterRef stays null.
 
     updateGlobalSettings({ radarEnabled: false });
 
-    // Nothing to assert on the adapter (we didn't create one); just
-    // verify no throw and that subsequent reads stay at the default.
+    // No throw; cache stays at the schema default.
     expect((getGlobalSettings() as Record<string, unknown>).radarEnabled).toBe(true);
-    expect(logger.warn).not.toHaveBeenCalled(); // logger isn't wired when init never ran
   });
 });
--- a/packages/deck-core/src/global-settings.test.ts
+++ b/packages/deck-core/src/global-settings.test.ts
@@ -1,0 +1,155 @@
+import type { ILogger } from "@iracedeck/logger";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  _resetGlobalSettings,
+  getGlobalSettings,
+  initGlobalSettings,
+  onGlobalSettingsChange,
+  updateGlobalSettings,
+} from "./global-settings.js";
+import type { IDeckPlatformAdapter } from "./types.js";
+
+type EchoCallback = (settings: unknown) => void;
+
+function createMockLogger(): ILogger {
+  return {
+    trace: vi.fn(),
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  } as unknown as ILogger;
+}
+
+interface MockAdapter {
+  adapter: IDeckPlatformAdapter;
+  echo: EchoCallback | null;
+  setGlobalSettings: ReturnType<typeof vi.fn<(settings: Record<string, unknown>) => void>>;
+  getGlobalSettings: ReturnType<typeof vi.fn<() => void>>;
+}
+
+function createMockAdapter(): MockAdapter {
+  const echoHolder: { echo: EchoCallback | null } = { echo: null };
+  const setGlobalSettings = vi.fn<(settings: Record<string, unknown>) => void>();
+  const getGlobalSettings = vi.fn<() => void>();
+
+  const adapter = {
+    onDidReceiveGlobalSettings: (cb: EchoCallback) => {
+      echoHolder.echo = cb;
+    },
+    setGlobalSettings,
+    getGlobalSettings,
+  } as unknown as IDeckPlatformAdapter;
+
+  return {
+    adapter,
+    get echo() {
+      return echoHolder.echo;
+    },
+    setGlobalSettings,
+    getGlobalSettings,
+  };
+}
+
+describe("global-settings cache (synchronous update on local writes)", () => {
+  let mock: MockAdapter;
+
+  beforeEach(() => {
+    _resetGlobalSettings();
+    mock = createMockAdapter();
+    initGlobalSettings(mock.adapter, createMockLogger());
+  });
+
+  it("updateGlobalSettings reflects the new value on the very next getGlobalSettings call", () => {
+    updateGlobalSettings({ radarEnabled: false });
+
+    expect((getGlobalSettings() as Record<string, unknown>).radarEnabled).toBe(false);
+  });
+
+  it("alternates radarEnabled across consecutive local writes without any host echo", () => {
+    updateGlobalSettings({ radarEnabled: false });
+    expect((getGlobalSettings() as Record<string, unknown>).radarEnabled).toBe(false);
+
+    updateGlobalSettings({ radarEnabled: true });
+    expect((getGlobalSettings() as Record<string, unknown>).radarEnabled).toBe(true);
+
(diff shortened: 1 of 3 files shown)


niklam:
@coderabbitai review

niklam:
@coderabbitai review

niklam:
@coderabbitai review

niklam:
@coderabbitai review

niklam:
@coderabbitai review

niklam:
@coderabbitai review

niklam:
@coderabbitai review

Issue #436, opened 2026-04-25
Lazy-load mock-impl/mock-data on non-Windows so they don't ship in the Windows pack


niklam:
## Background

Both native packages ship a mock implementation that is only ever executed on non-Windows platforms (used by developers on macOS/Linux, per `.claude/skills/cross-platform-development/SKILL.md`):

- `packages/audio-native/src/mock-impl.ts`
- `packages/iracing-native/src/mock-impl.ts` + `packages/iracing-native/src/mock-data/{session-info,snapshots,telemetry}.ts`

The Stream Deck and Mirabox plugins, however, only run on Windows (the Stream Deck install targets Windows; iRacing itself is Windows-only).

## Problem

Despite never executing on the user's machine, all of these files end up inside the packed `.streamDeckPlugin`:

```text
bin/node_modules/@iracedeck/iracing-native/dist/mock-impl.js
bin/node_modules/@iracedeck/iracing-native/dist/mock-data/session-info.js
bin/node_modules/@iracedeck/iracing-native/dist/mock-data/snapshots.js
bin/node_modules/@iracedeck/iracing-native/dist/mock-data/telemetry.js
bin/node_modules/@iracedeck/audio-native/dist/mock-impl.js
(plus all corresponding .d.ts and .d.ts.map files)
```

Why they ship today: the top-level `index.ts` of each native package re-exports the mock implementation, e.g.

```typescript
// packages/audio-native/src/index.ts
import { AudioNativeMock } from "./mock-impl.js";
export { AudioNativeMock } from "./mock-impl.js";
```

and `mock-impl.ts` itself top-level-imports its data:

```typescript
// packages/iracing-native/src/mock-impl.ts
import { MOCK_SESSION_INFO_YAML } from "./mock-data/session-info.js";
import { MOCK_SNAPSHOTS } from "./mock-data/snapshots.js";
import { buildTelemetryBuffer, getBufferSize, MOCK_VAR_HEADERS, MOCK_VAR_INDEX_MAP } from "./mock-data/telemetry.js";
```

ESM resolves top-level imports at module load time, so even on Windows where the mock branch is never taken, the files have to exist on disk for the plugin to load. They cannot be removed via `.sdignore` without breaking startup.

## Recommended approach

Refactor the platform-selection logic in both native packages so that the mock module is loaded via a dynamic `import()` inside the non-Windows branch:

```typescript
// packages/audio-native/src/index.ts (sketch)
export class AudioNative {
[...]

Issue #435, opened 2026-04-25
release-it bumps an incomplete subset of workspace packages


niklam:
## Problem

`scripts/release-hooks.mjs` bumps versions for a **hardcoded subset** of `packages/*/package.json` files. New packages added since the list was last touched are silently skipped on every release, so their `version` fields drift far behind the rest of the monorepo.

Current state from `HEAD` (master @ `4554fe9e`):

| Package | Version | In release-hooks list? |
|---------|---------|------------------------|
| `@iracedeck/audio-assets` | `1.14.0` | ❌ MISSING |
| `@iracedeck/audio-native` | `1.14.0` | ❌ MISSING |
| `@iracedeck/audio-scenarios` | `1.14.0` | ❌ MISSING |
| `@iracedeck/audio-service` | `1.14.0` | ❌ MISSING |
| `@iracedeck/event-bus` | `1.14.0` | ❌ MISSING |
| `@iracedeck/icon-composer` | `1.11.0` | ❌ MISSING (4 minor versions behind) |
| `@iracedeck/pi-components` | `1.13.1` | ❌ MISSING |
| `@iracedeck/sim-events-iracing` | `1.14.0` | ❌ MISSING |
| `@iracedeck/deck-adapter-elgato` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/deck-adapter-mirabox` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/deck-core` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/icons` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/iracing-actions` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/iracing-native` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/iracing-plugin-mirabox` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/iracing-plugin-stream-deck` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/iracing-sdk` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/logger` | `1.15.0-alpha.0` | ✅ |
| `@iracedeck/website` | `1.15.0-alpha.0` | ✅ |

8 of 19 workspace packages are not bumped on release.

## Root cause

`scripts/release-hooks.mjs` has a static `packageJsonPaths` array (lines 14–26) that has not been updated as new workspace packages were added.

## Recommended fix

[...]

PR #437, opened 2026-04-25
fix(build): clear Win32 MAX_PATH on installed plugin

niklam:
## Related Issue

N/A — discovered during a build-output audit, not from a pre-existing issue. Two follow-up issues were filed during the work: #435 (release-it bumps an incomplete subset of workspace packages) and #436 (lazy-load mock-impl/mock-data on non-Windows so they don't ship in the Windows pack).

## What changed?

Three commits that together drop the longest path inside the packaged `.streamDeckPlugin` from **128 chars to ~76 chars**, clearing the 260-char Win32 `MAX_PATH` limit with comfortable headroom (and shrinking the install at the same time).

### Commit 1 — `fix(build): shorten installed plugin paths to clear Win32 MAX_PATH` (`f59c6161`)

- Move `node-addon-api` from `dependencies` → `devDependencies` in both `@iracedeck/audio-native` and `@iracedeck/iracing-native`. It is a build-time-only set of C++ headers consumed by `binding.gyp` (`require('node-addon-api').include`); the compiled `.node` artefact does not need it at runtime, so it shouldn't ship in the installed plugin's `node_modules`. This eliminates the deepest `node-addon-api/` subtree.
- Bundle `@iracedeck/audio-service` into `plugin.js` for both the Stream Deck and Mirabox plugins (drop it from the `external` array and from the emitted `bin/package.json`). audio-service is pure TypeScript with audio-native as its only runtime peer (which stays external), so rollup can inline it the same way it inlines deck-core. This eliminates the entire `bin/node_modules/@iracedeck/audio-service/` subtree, which previously contained a *duplicated nested copy* of audio-native + node-addon-api.

### Commit 2 — `fix(build): drop unused native-package files from the packed plugin` (`0dada74a`)

Add `.sdignore` rules (identical for both plugins) that match each native package's published `files` whitelist in `package.json`. `file:` deps don't honor `files`, so without these rules the entire package directory ships:

- `**/node_modules/@iracedeck/*/assets/` — kills the iRacing SDK reference projects (~413 KB; previously the source of the 100-char paths).
- `**/node_modules/@iracedeck/*/scripts/` and `README.md` — never read at runtime.
- `**/build/{*.sln,*.gypi,*.vcxproj,*.vcxproj.filters}` — only the `.node` binary in `build/Release/` is needed.
- `**/node_modules/@iracedeck/*/Python/` — defensive, catches a node-gyp Python tree that can appear in the pre-pack tree under some build states.

### Commit 3 — `fix(build): drop compiled test artifacts from packed plugin dist/` (`1f20659b`)

Add `**/node_modules/**/dist/**/*.test.*`. Strips 9 `.test.js` / `.test.d.ts` / `.test.d.ts.map` files (the longest 3 paths after commit 2). `mock-data/*.{js,d.ts}` are intentionally retained — `mock-impl.ts` top-level-imports them, so they must resolve at module-load time on every platform. See #436 for a follow-up that would let these be dropped too via a code refactor.

### Path-length progression (verified against extracted `.streamDeckPlugin`)

| Build | Longest relative path | Top culprit |
|---|---|---|
| Original | 128 chars | `audio-service/node_modules/@iracedeck/audio-native/node_modules/node-addon-api/...` |
| After commit 1 | 100 chars | `iracing-native/assets/irsdk_1_19/.../*.vcxproj` |
| After commit 2 | 81 chars | `iracing-native/dist/mock-data/telemetry.test.d.ts.map` |
| After commit 3 | ~76 chars | The `.node` binary itself (`build/Release/iracing_native.node`) — intentional |

Headroom on a typical install (`%APPDATA%\Elgato\StreamDeck\Plugins\com.iracedeck.sd.core.sdPlugin\` ≈ 87 chars): **~97 chars of slack** under MAX_PATH; ~75 chars even with a long-username + OneDrive prefix.

## How to test

- [ ] `pnpm build` succeeds for both plugins.
- [ ] Pack each plugin (e.g. via the existing GitHub Action), extract the resulting `.streamDeckPlugin`, and recursively measure file paths. Confirm:
  - No `bin/node_modules/@iracedeck/audio-service/` subtree.
  - No `bin/node_modules/@iracedeck/audio-native/node_modules/` subtree.
  - No `node_modules/.../assets/irsdk_1_19/...` paths.
  - No `dist/**/*.test.*` files.
  - Longest path ≤ ~80 chars.
- [ ] Install the rebuilt Stream Deck plugin and trigger a Pit Crew action (exercises bundled audio-service → external audio-native at runtime). Confirm audio plays.
- [ ] `pnpm test` passes across the workspace.

## Checklist

- [ ] Linked to an approved issue *(N/A — internal build-output finding; see Related Issue)*
- [x] New code has unit tests *(N/A — build/packaging configuration only, no new code paths)*

## Summary by CodeRabbit

* **Chores**
  * Moved native helper package to development-only to reduce runtime installs.
  * Expanded packaging ignores to exclude build artifacts and bundled non-runtime files, slimming plugin packages.
* **Build / Packaging**
  * Plugins now bundle the audio service instead of treating it as an external dependency, so the service is included in plugin installs.

### DIFF
--- a/packages/audio-native/package.json
+++ b/packages/audio-native/package.json
@@ -18,11 +18,9 @@
     "src/*.cc",
     "src/*.h"
   ],
-  "dependencies": {
-    "node-addon-api": "8.7.0"
-  },
   "devDependencies": {
     "@types/node": "25.6.0",
+    "node-addon-api": "8.7.0",
     "node-gyp": "12.2.0",
     "rimraf": "6.1.3",
     "typescript": "5.9.3",
--- a/packages/iracing-native/package.json
+++ b/packages/iracing-native/package.json
@@ -17,11 +17,9 @@
     "build/Release/*.node",
     "src/*.cc"
   ],
-  "dependencies": {
-    "node-addon-api": "8.7.0"
-  },
   "devDependencies": {
     "@types/node": "25.6.0",
+    "node-addon-api": "8.7.0",
     "node-gyp": "12.2.0",
     "rimraf": "6.1.3",
     "typescript": "5.9.3",
--- a/packages/iracing-plugin-mirabox/com.iracedeck.sd.core.sdPlugin/.sdignore
+++ b/packages/iracing-plugin-mirabox/com.iracedeck.sd.core.sdPlugin/.sdignore
@@ -15,8 +15,17 @@
 # MSBuild tracking logs
 *.tlog
 
+# node-gyp generated build files (only build/Release/*.node is needed at runtime)
+**/build/*.sln
+**/build/*.gypi
+**/build/*.vcxproj
+**/build/*.vcxproj.filters
+
 # Python bytecode cache (node-gyp)
 **/__pycache__/
+
+# Standalone Python distribution downloaded by node-gyp
+**/node_modules/@iracedeck/*/Python/
 
 # Dev dependencies (nested in native modules)
 **/node_modules/typescript/
@@ -28,6 +37,11 @@
 # Source files (already compiled)
 **/node_modules/*/src/
 **/node_modules/@*/*/src/
+
+# iRacing SDK reference projects and build scripts (not runtime artifacts)
+**/node_modules/@iracedeck/*/assets/
+**/node_modules/@iracedeck/*/scripts/
+**/node_modules/@iracedeck/*/README.md
 
 # Dev/build config files
 **/tsconfig.json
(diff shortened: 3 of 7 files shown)


niklam:
@coderabbitai applied your nitpick on the `.sdignore` `@iracedeck/*/` patterns in commit fa3892d6.

The wildcard rules for `assets/`, `scripts/`, `README.md`, and `Python/` are now listed explicitly per-package for `audio-native` and `iracing-native`. Same logic applied symmetrically to both plugins. A short comment in the file explains why it's per-package on purpose, so a future maintainer adding a third native package has to make an explicit decision about each pattern instead of inheriting the silent-drop behavior the wildcard would have caused.

niklam:
@coderabbitai review

PR #461, opened 2026-04-26
chore(audio-scenarios): delete dead Legacy* / PENDING_TOGGLE_SCENARIOS block (#460)

niklam:
## Related Issue

Fixes #460

## What changed?

The \"Pending migration (NOT registered)\" block in `packages/audio-scenarios/src/catalog/pit-crew/toggle-confirmations.ts` was kept as a template for future windshield / fastRepair / DRS / P2P migration PRs. It is dead code today:

- Nothing imports `PENDING_TOGGLE_SCENARIOS` or any of the legacy helpers (`legacyPitServiceScenario`, `legacyCarControlScenario`, `LEGACY_PIT_SERVICE_CLIP`, `PitServiceLegacy`) outside this file.
- Its clip paths still reference the deleted `pit-crew/` asset tree, so the scenarios can't run as-is.
- The registered `fuelScenario` / `tireSetOnScenario` / `compoundScenario` already demonstrate the modern pattern; any future toggle confirmation will copy them rather than the legacy template.

This PR:

- `packages/audio-scenarios/src/catalog/pit-crew/toggle-confirmations.ts` — drops the legacy block (~70 lines), prunes the header doc comment that advertised the removed export.
- `docs/superpowers/specs/2026-04-26-flag-voice-scenarios-design.md` — removes the \"Out of scope\" bullet that named `PENDING_TOGGLE_SCENARIOS` as a separate follow-up.

No production behavior change — the block was never registered.

## How to test

- [x] `pnpm test` — 3389 / 3389 pass (same count as master).
- [x] `pnpm lint` — clean.
- [x] `pnpm build` — 17/17 turbo tasks succeed.
- [x] `grep -R PENDING_TOGGLE_SCENARIOS packages/ docs/` — no remaining references.

## Checklist

- [x] Linked to an approved issue
- [x] New code has unit tests _(deletion only — covered by the existing 3389-test suite continuing to pass)_
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) _(audio-scenarios is shared; both plugins consume the same scenario catalog)_
- [x] No unrelated changes included

## Summary by CodeRabbit

* **Chores**
  * Removed pending toggle scenario implementations and updated related documentation. Code cleanup with no impact to user-facing functionality.

PR #462, opened 2026-04-26
fix(pit-crew): default Race Engineer and Radar to off (#378)

### LINKED ISSUE #378: fix(pit-engineer): default to off and promote on/off to global setting
## What would you like?

Two related changes to how the Pit Engineer master toggle behaves:

1. **Default state must be off.** A fresh install or a newly-added button currently starts with the engineer active — that's intrusive for users who don't want voice callouts. Flip the default so users have to opt in.
2. **Master on/off must be a global setting, not per-action.** Today the toggle is stored per action instance. If a user adds two Pit Engineer buttons, they can end up in contradictory states. The engineer is a plugin-wide service; its on/off belongs in global settings so every instance reflects the same state and every instance's button shows the same icon.

## Use case

- Onboarding: first launch should be quiet. Users discover the feature, enable it intentionally.
- Multiple buttons: some users bind one button on the Stream Deck and another on a touch deck / Mirabox. Those should agree on "engineer is on."
- Honest UX: the state is plugin-global in behavior (it gates all telemetry subscriptions), so storing it as global settings matches reality.

## Area

Action (existing)

## Alternatives considered

- Ship default-on with an onboarding modal: more friction than just defaulting off.
- Keep per-action state and sync it manually across instances: fragile; breaks when one instance isn't visible.

## Additional context

- Branch: `feature/pit-engineer` (part of #375).
- Related to the master toggle stored inside `PitEngineerSettings` on the branch. The sub-feature checkboxes can stay per-action (user may want one button with flags only, another with pit callouts), but the master on/off is plugin-wide.
- When this is wired, the button icon should reflect the global state on all visible instances.

niklam:
## Related Issue

Fixes #378

## What changed?

Both `raceEngineerEnabled` and `radarEnabled` previously defaulted to `true`, so a fresh install — or any user who had never pressed either toggle — started with the engineer voice and proximity ticks already running. Per issue #378, the master gate(s) must default **off** so onboarding is quiet and users explicitly opt in. The user requested the same change for Radar (same UX problem), so this PR flips both.

### Schema (`packages/deck-core/src/global-settings.ts`)

- `raceEngineerEnabled: ...default(false)`
- `radarEnabled: ...default(false)`
- Doc comments updated with the rationale and the issue link.

### Action helpers (`packages/iracing-actions/src/actions/pit-crew/pit-crew.ts`)

- `isRaceEngineerEnabled()` and `isRadarEnabled()` now use `=== true` instead of `!== false`, so an unset cache reads as off (matching the schema default and #378's \"fresh install stays quiet\" intent).

### Scenario harness (`packages/scenario-harness/src/main.ts`)

- `applyAudioSettings` now uses the same `=== true` semantic so the harness behaves identically to the shipped plugin. The harness bootstrap still seeds both flags as true so demo runs start with audio audible — that's intentional and unchanged.

### Tests (`packages/deck-core/src/global-settings.test.ts`)

- `read → flip → write → read` simulation rewritten for the new helper semantic: first read on the unset cache is now `false` → flips → writes `true`, and the alternation continues from there.
- The Zod-default merge test now expects `raceEngineerEnabled: false`.
- The \"no-ops when adapter is not initialized\" test seeds `radarEnabled: true` and asserts the cached default stays at `false`.

### Docs

- `docs/plugins/core/actions/pit-crew.md` — new \"Default state\" section spelling out that both gates ship off and that the icon paints red until the user opts in.
- `packages/website/src/content/docs/docs/actions/driving/pit-crew.md` — top-level note about default-off behavior plus per-mode hints.

## How to test

- [x] `pnpm test` — 3389 / 3389 pass.
- [x] `pnpm lint` — clean.
- [x] `pnpm build` — 17/17 turbo tasks succeed.
- [ ] **Manual** (post-merge, fresh install): drop a Pit Crew \"Race Engineer Toggle\" key onto a fresh Stream Deck profile. Expected: the icon's status bar paints red on first load, and pit-service / flag callouts stay silent. Press once: status bar flips green and the engineer audio resumes. Repeat for the Radar mode (status bar red until first press, ticks silent until then).

## Checklist

- [x] New code has unit tests _(updated existing global-settings tests; the pit-crew suite already covers both enabled / disabled bus states)_
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) _(action lives in shared `@iracedeck/iracing-actions`; both plugins share the deck-core schema)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Race Engineer and Radar features now default to disabled on fresh installs, requiring explicit key press activation to enable. System starts in quiet mode until user enables these features.

* **Documentation**
  * Updated documentation to clarify that Race Engineer and Radar modes start disabled by default, with enabled state synchronized globally across all Pit Crew buttons.

### DIFF
--- a/docs/plugins/core/actions/pit-crew.md
+++ b/docs/plugins/core/actions/pit-crew.md
@@ -11,11 +11,15 @@ Multi-mode action covering the iRaceDeck pit-side audio framework. Modes availab
 | SDK Support | No |
 | Encoder Support | No |
 
+## Default state
+
+Both `raceEngineerEnabled` and `radarEnabled` ship **off** (issue #378). A fresh install — and any user who has never pressed either toggle — stays quiet until they explicitly enable the feature with a Race Engineer Toggle or Radar key press. The status bar on each toggle's icon paints red on first launch, flipping green only after the first press.
+
 ## Behavior
 
 ### Button Press
-- **Race Engineer Toggle mode**: Flips the plugin-global `raceEngineerEnabled` gate. When off, both `AudioBus.Voice` (engineer messages, acks, toggle confirmations) and `AudioBus.Background` (pit ambient loop and walkie-talkie SFX) are zeroed synchronously, so any in-flight clip silences on the same key press. `AudioBus.Alerts` (radar) is intentionally untouched — radar has its own toggle. Re-enabling restores Voice to the configured `Race Engineer Volume` and Background to unity.
-- **Radar mode**: Flips the plugin-global `radarEnabled` and stops/starts the directional proximity tick loop synchronously. Used by Radar alongside the per-instance Radar Test button.
+- **Race Engineer Toggle mode**: Flips the plugin-global `raceEngineerEnabled` gate. When off (the default), both `AudioBus.Voice` (engineer messages, acks, toggle confirmations) and `AudioBus.Background` (pit ambient loop and walkie-talkie SFX) are zeroed synchronously, so any in-flight clip silences on the same key press. `AudioBus.Alerts` (radar) is intentionally untouched — radar has its own toggle. Re-enabling restores Voice to the configured `Race Engineer Volume` and Background to unity.
+- **Radar mode**: Flips the plugin-global `radarEnabled` and stops/starts the directional proximity tick loop synchronously. Off by default — pressing the key once starts the loop. Used by Radar alongside the per-instance Radar Test button.
 - **Radar Volume mode**: Steps the plugin-global `radarVolume` by ±5, clamped to 0–100. Takes effect immediately on `AudioBus.Alerts`. Direction is configured per button (Up or Down). Stepping to 0 mutes the radar without toggling the feature off.
 
 ### Race Engineer voice coverage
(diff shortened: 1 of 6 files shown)


niklam:
**Re: Docstring Coverage pre-merge warning (25.00% vs 80.00% threshold)**

Declining, same reasoning as on #461.

The substantive surface in this PR is:
- two `.default(true) → .default(false)` literals in a Zod schema (the surrounding JSDoc blocks already document each field's semantics, and they're updated in this commit),
- two private one-line helpers `isRaceEngineerEnabled` / `isRadarEnabled` that became `=== true` reads (the first carries a block comment explaining why `=== true` rather than `!== false`; the second is a literal mirror of it),
- one arrow function in `applyAudioSettings` for parity with the production helper,
- test edits and Markdown doc updates.

Adding docstrings to symbols whose bodies and surrounding context already explain them — or to test-suite functions, which the project doesn't doc-comment — would be noise and violates `.claude/rules/code-style.md`'s "default to writing no comments / only add one when the WHY is non-obvious" guideline. The non-obvious WHY (`=== true` vs `!== false`) is already documented inline.

The four substantive pre-merge checks (Title / Description / Linked Issues / Out of Scope) and every status check (Prettier, ESLint, Run tests, CodeRabbit, PR Labeler) are green. Merging.

PR #475, opened 2026-04-28
feat(audio-scenarios): per-callout opt-in/out controls (#467)

### LINKED ISSUE #467: feat(audio-scenarios): per-flag-color opt-in/out controls
## What would you like?

Per-flag-color opt-in/out controls so users can disable individual Race Engineer flag callouts they find noisy or irrelevant. Most prominent example: the **debris flag** can re-fire on every on→off→on cycle of `Flags.Debris` with no cooldown, which a tester described as "REALLY repetitive".

Concrete shape:

- A new "Race Engineer flag callouts" accordion in the Pit Crew Property Inspector (or in plugin-global settings — see Alternatives), with one **enabled** checkbox per flag color. All enabled by default to preserve current behavior.
- Settings stored as **plugin-global** so the choice is shared across every Pit Crew button.
- Disabled flags simply don't register their scenario at startup (the underlying sim event still fires; only the audio scenario is suppressed).

Initial set of toggles, one per flag scenario shipping today (11 total):

`flag-yellow-local`, `flag-yellow-full`, `flag-yellow-cleared`, `flag-green`, `flag-blue`, `flag-white`, `flag-red`, `flag-black`, `flag-checkered`, `flag-debris`, `flag-meatball`.

## Use case

Driver feedback (projektdotnet42, Discord): _"Need ability to disable flags (either all or specific ones, your call). The debris flag gets REALLY repetitive."_

The Race Engineer voice pipeline is one of the project's flagship features for VR / heads-down drivers, but a single noisy flag can sour the whole experience and push users to disable Race Engineer entirely. Giving users a granular kill-switch keeps the rest of the callouts useful.

## Area

Action (existing) — `Pit Crew` (`com.iracedeck.sd.core.pit-crew`), Race Engineer voice family. Affects the `@iracedeck/audio-scenarios` Pit Crew flag catalog and `@iracedeck/deck-core` global settings.

## Alternatives considered

- **A single "disable all flag callouts" master toggle.** Trivial to implement but doesn't solve the user's actual problem — they want to keep most flags. Rejected.
- **Per-action settings instead of plugin-global.** Inconsistent with the rest of the Race Engineer settings (voice, volume, enabled) which are already global, and would require users to repeat the choice on every Pit Crew button. Rejected.
- **A debris-specific cooldown** (e.g., suppress within 30 s of the last fire). Helps the noisy case but doesn't generalize, and a user who simply doesn't care about debris still gets it. Cooldown could be a follow-up that applies to *all* flag scenarios uniformly.
- **PI multi-select dropdown instead of N checkboxes.** Less intuitive; the checkbox grid is the standard Property Inspector idiom and matches what users expect from an "enable / disable" list.

### Required code changes

**Global settings (`@iracedeck/deck-core`)**

- `packages/deck-core/src/global-settings.ts` — add boolean keys per flag, all defaulting to `true`. Naming convention: `flagCalloutEnabled<Color>` (e.g., `flagCalloutEnabledDebris`, `flagCalloutEnabledYellowLocal`). Add to `GlobalSettingsSchema`.

**Scenario registration (`@iracedeck/audio-scenarios`)**

- `packages/audio-scenarios/src/catalog/pit-crew/index.ts` — `registerPitCrew(bus)` already iterates the flag scenarios from `flag-alerts.ts`. Filter the registration list by the corresponding global setting. The filter runs at startup; settings changes after that take effect on plugin reload (acceptable; matches how voice changes are picked up today).
- No changes to `flag-alerts.ts` itself — the scenarios stay defined; the registration step is what gates them.

**Property Inspector (`@iracedeck/iracing-actions`)**

- `packages/iracing-actions/src/actions/pit-crew/pit-crew.ejs` — add a new accordion ("Flag callouts") with one `<sdpi-checkbox setting="flagCalloutEnabled<Color>" global default="true">` per flag. Place under the existing Race Engineer settings section.

**Tests**

- `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.test.ts` — add coverage that disabled flags are not registered with the engine.
- Schema test for the new global setting defaults.

### Cross-cutting artifacts (per `.claude/rules/build-and-commit.md`)

- **Plugin manifests** — no change (no new actions or modes).
- **Website** — `packages/website/src/content/docs/docs/actions/driving/pit-crew.md`: document the new flag-callout toggles per `.claude/rules/website-action-docs.md`.
- **Skills** — none required (no action / mode / sub-action count changes).
- **Rules** — `.claude/rules/global-settings.md` if the new settings deserve mention as a precedent for "list of toggles" patterns.

### Out of scope

- A flag-callout cooldown / rate-limit. Separate follow-up if "I want fewer of these, not zero" turns out to be the more common ask.
- Live re-registration when settings change mid-session. Plugin reload is good enough for a v1.
- Per-voice variation of which flags are enabled. Settings apply to whichever voice is active.
- Disabling non-flag callouts (pit-service toggles, limiter warnings, etc.) — those have their own usage patterns and warrant a separate design.

### Related

- #463 — blue-flag reminder loop and leader-aware variants. Both this issue and #463 touch `flag-alerts.ts` and the per-flag pool catalog; whichever lands first should keep the registration shape simple for the other.

niklam:
## Related Issue

Fixes #467

## What changed?

Per-callout opt-in/out for the Race Engineer voice family, starting with all 11 flag colours, plus a project-wide naming convention for future families.

- **Storage:** one boolean global setting per callout subject, named `callout<Polarity><Family><Subject>` (e.g. `calloutEnabledFlagDebris`). All 11 flag toggles default `true`. A grep for `calloutEnabled` finds every callout-toggle setting in the codebase. Per-item booleans (rather than an array or bitmask) are the load-bearing choice for forward-compat: a future newly-added subject's Zod default flows to every existing user via `passthrough()`, so adding a new callout never silently disables it for anyone.
- **Live mid-session gating** (in scope despite #467 marking it for v1+): each flag scenario's `where:` predicate reads the live `getGlobalSettings()` cache on every event arrival, BEFORE `attemptFire`. Toggling a flag off does **not** cut a callout already playing — only future events of that subject are suppressed. No `onGlobalSettingsChange` listener; no re-registration churn.
- **PI:** new "Race Engineer Callouts" accordion (closed by default) in the Pit Crew Property Inspector. Single `sdpi-item label="Flags"` containing 11 stacked labelled checkboxes laid out in a 2-column CSS grid, computed from the list length so the layout stays at exactly two columns regardless of how the family grows (commit 2).
- **Plugin wiring:** both `iracing-plugin-stream-deck` and `iracing-plugin-mirabox` pass a `getGlobalSettings()`-reading closure into `registerPitCrew`. Init order unchanged; Zod defaults seed the cache before any event fires.
- **Documentation:** website `pit-crew.md` + `.claude/rules/global-settings.md` document the convention and the array-driven 2-column grid pattern as project precedent.
- **Follow-up tracked separately:** ship a custom `<ird-checkbox-list>` component when the second family lands (Pit Actions) so it has two consumers at once. The persistence shape (per-item booleans) doesn't change.

Tests: new `register-pit-crew.test.ts` covers disable → no fire, mid-clip toggle off → in-flight clip completes, re-enable → next fires, scope predicates still respected, meatball-disabled doesn't preempt, debug-log fires on suppression. Schema defaults + boolean/string coercion covered for all 11 keys.

Builds: `pnpm build`, `pnpm test` (3479 passing), `pnpm lint:fix`, `pnpm format:fix` all clean from the worktree.

## How to test

- [x] Stream Deck: load Pit Crew → Race Engineer mode. Open the Property Inspector. The new "Race Engineer Callouts" accordion is closed by default; opening it shows a single "Flags" item with 11 checkboxes laid out in 2 columns (6 + 5), all checked.
- [x] Trigger a flag event in a session — confirm the engineer announces it.
- [x] Uncheck **Debris**. Repeatedly trigger debris on→off→on — confirm no audio plays for debris while other flags still announce.
- [x] With debris re-checked, start a long callout (e.g. yellow-full), uncheck **Yellow (full course)** *while it is playing*. Confirm playback completes naturally; no cut. Trigger another yellow-full event — confirm it is suppressed.
- [ ] Re-check **Yellow (full course)** — next event fires the callout again.
- [ ] Repeat smoke-test on the Mirabox plugin.
- [ ] Forward-compat sanity: hand-edit the persisted global-settings JSON to remove every `calloutEnabledFlag*` key, relaunch — all 11 boxes should render checked (proves Zod schema default + `default="true"` work together).

## Checklist

- [x] New code has unit tests

## Summary by CodeRabbit

* **New Features**
  * Added per-flag callout toggles in Race Engineer settings, allowing users to individually enable/disable each flag callout subject.
  * Toggles apply live during a session; newly suppressed callouts won't play, but already-playing audio clips complete.
  * Disabling a callout also disables its preemption capability.

* **Documentation**
  * Added documentation for the new callout toggle feature.


niklam:
@coderabbitai help

PR #487, opened 2026-05-01
fix(sim-events-iracing): clear-tires no longer mis-fires compound callout (#484)

### LINKED ISSUE #484: fix(audio-scenarios): clearing tires re-announces the compound and skips the "tires cleared" line
### What happened?

With a non-default compound queued (engineer has already said e.g. "switching to wet"), pressing "clear tires" causes the engineer to say "switching to dry" again instead of acknowledging the tire clear. The "tires cleared" / "skip tires" line never plays.

If the same "clear tires" press happens with the car's default compound already selected, the bug doesn't appear — the clear ack plays normally. The bug needs a non-default compound to be queued.

### Expected behavior

Pressing "clear tires" should produce the "tires cleared" / "skip tires" voice line regardless of which compound was previously selected. Re-announcing the compound when the user explicitly cleared all tire service is misleading.

### Steps to reproduce

1. In iRacing, queue a tire compound that differs from the car's default (e.g. switch to wet on a car that defaults to dry).
2. Hear the engineer say "switching to wet" — correct.
3. Press "clear tires" (or "clear all pit service").
4. Hear the engineer say "switching to dry" — wrong (no tire clear ack).

### Additional context

**Root cause (one verification step still needed in-sim).**

`packages/sim-events-iracing/src/diff/toggles.ts:179-204` deliberately suppresses the cascading tire-set diff whenever the compound flag flips in the same tick:

\`\`\`ts
const baselineTireBits = compoundJustChanged ? currTireBits : state.lastPitSvFlags & TIRE_FLAGS_MASK;
// ...
if (compoundJustChanged) {
  state.lastSeenTireFlags = currTireBits;
  state.lastTireChangeAt = 0;
}
\`\`\`

The intent (per the comment at lines 179-184) is correct for genuine user-initiated compound flips: iRacing force-sets all four tire bits in the same tick, and we want one canonical "switching to wet" line, not "switching to wet" plus "all four tires" 500 ms later.

The bug appears when "clear tires" causes iRacing to reset the compound bit to its car-default in the same tick as the tire bits flip to zero. Suspected sequence:

1. iRacing's clear path resets the compound bit back to the car default (likely 0 / dry).
2. Translator emits \`tireService.compoundChanged { from: 1, to: 0 }\`.
3. The matching compound scenario fires: "switching to dry" — what the user hears.
4. The same translator branch that absorbs the cascade for genuine compound flips also suppresses the legitimate \`tireService.changed { current: [] }\` event — so \`TIRE_OFF_SCENARIO\` ("tires cleared", \`packages/audio-scenarios/src/catalog/pit-crew/toggle-confirmations.ts:152-164\`) never fires.

The suppression rule does not distinguish "user actively switched compound (cascade should be absorbed)" from "compound was reset as a side-effect of clearing pit service (the clear is the meaningful announcement)". The two cases differ at the data level — \`currTireBits === 0\` for the clear case versus \`currTireBits === TIRE_FLAGS_MASK\` for the compound flip — but the current code does not look at \`currTireBits\` when deciding whether to suppress.

**Verification still owed before fix.**

Capture \`PitSvFlags\` and \`PitSvCompound\` across the actual "clear tires while non-default compound queued" sequence and confirm the bit pattern. That rules out alternative explanations (e.g. a race between two compound events).

### Related

- #468 — windshield + fast-repair toggle acknowledgements (separate, related Race Engineer ack gap).
- #481 — pit-readback replays with stale snapshot (separate, related Race Engineer state-management bug).

niklam:
## Related Issue

Fixes #484

## What changed?

Pressing "clear tires" while a non-default compound was queued caused
iRacing to reset the compound bit back to the car default in the same
tick that all four tire bits cleared. The translator treated that as a
genuine user-initiated compound flip and:

1. Emitted `tireService.compoundChanged` → engineer mis-spoke
   "switching to dry" instead of acknowledging the clear.
2. Suppressed the cascading `tireService.changed { current: [] }` under
   the same compound-flip absorption rule, so the legitimate "tires
   cleared" callout never fired.

The two cases differ at the data level: a real user-initiated compound
flip force-sets all four tire bits (`currTireBits === TIRE_FLAGS_MASK`);
the clear-tires side-effect leaves them at zero. The fix gates the
`tireService.compoundChanged` emit on the all-four-tires cascade so the
clear-tires path falls through to the regular tire-set diff and the
"tires cleared" callout fires.

Tests:
- Adds a translator test exercising the clear-tires-with-non-default-
  compound sequence end-to-end (compound-flip first, then clear, asserts
  exactly one compound event + one `tireService.changed { current: [] }`).
- Updates the existing dry↔wet compound tests to include the all-four-
  tires cascade so they reflect real iRacing behavior rather than the
  synthetic isolated-bit-flip case (which the new gating no longer
  honors).

## How to test

- `pnpm test` — 3556 tests pass, including the new #484 regression test.
- `pnpm build` — full workspace build green.
- Manual repro:
  1. In iRacing, queue a non-default tire compound (e.g. wet on a
     dry-default car). Engineer says "switching to wet".
  2. Press "clear tires" (or "clear all pit service").
  3. Engineer should now say "tires cleared" / "skip tires" — NOT
     "switching to dry".

## Checklist

- [x] New code has unit tests

### DIFF
--- a/packages/sim-events-iracing/src/diff/toggles.ts
+++ b/packages/sim-events-iracing/src/diff/toggles.ts
@@ -182,9 +182,18 @@ export function diffToggles(state: TranslatorState, telemetry: TelemetryData, no
   // absorb the cascading tire-set diff so the compound voice line is the
   // single canonical confirmation (otherwise the engineer would also call
   // out "all four tires" 500 ms later).
+  //
+  // Issue #484: a "clear tires" press also flips the compound bit (iRacing
+  // resets compound to the car default as a side-effect of clearing pit
+  // service) but with `currTireBits === 0` instead of `TIRE_FLAGS_MASK`.
+  // That isn't a user-initiated compound change — it's a consequence of
+  // the clear — so emitting "switching to dry" would be misleading and
+  // would also suppress the legitimate "tires cleared" callout. Gate the
+  // compound emit on the all-four-tires cascade so only genuine user
+  // compound flips fire the compound voice line.
   let compoundJustChanged = false;
 
-  if (state.lastPitSvCompound !== pitSvCompound) {
+  if (state.lastPitSvCompound !== pitSvCompound && currTireBits === TIRE_FLAGS_MASK) {
     emit({
       event: "tireService.compoundChanged",
       data: { from: state.lastPitSvCompound, to: pitSvCompound },
--- a/packages/sim-events-iracing/src/translator.test.ts
+++ b/packages/sim-events-iracing/src/translator.test.ts
@@ -509,8 +509,16 @@ describe("sim-events-iracing translator", () => {
       bus.subscribe("tireService.compoundChanged", handler);
       initializeSimEventsIracing(bus, controller, createMockLogger());
 
-      controller.__tick(telemetry({ PitSvTireCompound: 0 }));
-      controller.__tick(telemetry({ PitSvTireCompound: 1 }));
+      const allTires =
+        PitSvFlags.LFTireChange | PitSvFlags.RFTireChange | PitSvFlags.LRTireChange | PitSvFlags.RRTireChange;
+
+      // iRacing always force-sets all four tire bits in the same tick as a
+      // user-initiated compound flip; the translator gates the
+      // compoundChanged emit on that cascade so an isolated compound bit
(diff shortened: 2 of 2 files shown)


niklam:
@coderabbitai review

Issue #471, opened 2026-04-27
feat(pit-crew): independent volume control for ambient/SFX background bus


niklam:
## What would you like?

A new **Background volume** control (0–100, default 100, step 1) in the Pit Crew Property Inspector that scales the `AudioBus.Background` output. The Background bus carries both pit/track ambient loops (channel 0, mix 0.8) and walkie-talkie SFX (channel 1, mix 0.7); a single slider adjusts both proportionally.

The setting is plugin-global (matching `raceEngineerVolume` and `radarVolume`), persists across sessions, and applies live without a reload.

## Use case

Driver feedback (projektdotnet42, Discord, on the Race Engineer pit ambience and radio SFX): _"They work pretty well, though maybe tune the background sounds down just a touch — they add realism but at the cost of usability for people like me who have minor audio processing issues (sometimes I just can't hear without my subtitles)."_

The user attributes the difficulty to ADHD plus drum-induced hearing damage; in practice this is a common accessibility ask — let users dial down ambient/SFX so the engineer voice cuts through. Today there's only one knob (`raceEngineerVolume`) which controls the voice bus; turning it up makes the voice louder *and* the background louder relative to system volume.

## Area

Action (existing) — `Pit Crew` (`com.iracedeck.sd.core.pit-crew`) and `@iracedeck/deck-core` global settings.

## Alternatives considered

- **Two separate sliders (Ambient + SFX).** Both currently route to the same `AudioBus.Background`. To control them independently we'd either split into two buses (`Ambient` + `SFX`) or expose channel-level mix-ratio adjustments. Both are ~3× the work for a v1; the user only asked for "tune background down a touch" and a single slider solves that fully. Independent control can be a follow-up if accessibility feedback warrants it.
- **A "background mute" checkbox** instead of a slider. Loses the "just a touch" granularity that was the user's actual ask. Rejected.
- **Hide the slider behind an "Accessibility" accordion** to avoid clutter. Standard volume knob, fits naturally next to the existing voice/radar sliders. Keep it visible.
- **Reuse the `radar-volume` stepper-mode pattern** (a Pit Crew sub-mode with up/down keys instead of a PI slider). The radar stepper exists because users want to tweak it during a session without opening PI; background level is a set-and-forget preference, so PI-only is fine. Add a stepper in a follow-up if the request comes in.

### What already exists

[...]

PR #523, opened 2026-05-07
fix(audio-assets): drop bake-in gain from radio filter; lower default volumes (#522)

### LINKED ISSUE #522: fix(audio-assets): drop bake-in gain from radio filter; lower default volumes (RE 50, Ambient 25, Radar 50)
## What would you like?

Two related volume fixes for the Race Engineer audio path:

### 1. Drop the bake-in makeup gain from the radio filter

`packages/audio-assets/src/presets.mjs` (`RADIO_ENGINEER_FILTER`) currently chains:

```text
highpass=f=250,lowpass=f=3500,volume=8dB,asoftclip=type=tanh,volume=-6dB,volume=6dB,alimiter=limit=0.95
```

Net result: **+8 dB of pre-gain** stamped into every voice clip at build time, riding the soft-clip and limiter. The clips ship hot, so the user's `raceEngineerVolume` slider has to be pulled way down before the voice sits at a sensible level next to game audio — the slider effectively wastes its top half undoing baked-in gain.

Drop the makeup so the filter does what it says on the tin (bandpass + gentle clip + brick-wall safety) without juicing the loudness. Concretely: remove the `volume=8dB` pre-gain and the `volume=-6dB,volume=6dB` cancelling pair, leaving the bandpass + soft-clip + limiter doing only their stated jobs. Re-render every voice clip.

Cache invalidation is automatic — the cache path embeds a hash of the filter chain, so the next build reprocesses every voice clip from source.

### 2. Lower default channel volumes

Update `packages/deck-core/src/global-settings.ts`:

| Setting | Current default | New default |
|---|---|---|
| `raceEngineerVolume` | 100 | **50** |
| `backgroundVolume` (Ambient + walkie SFX) | 35 | **25** |
| `radarVolume` | 100 | **50** |

These match a more reasonable out-of-the-box mix — pairs naturally with the un-baked-gain voice clips above and gives users headroom in both directions instead of always sliding everything down.

## Use case

Today the out-of-the-box Race Engineer experience is "instantly too loud" — the bake-in gain plus 100 % default makes the voice dominate game audio. Every new user has to find the volume sliders and lower them before the feature is usable. Lowering the defaults plus removing the hidden gain means the first-run experience already sits at a sensible level, and users with audio-processing sensitivities (already called out in the `backgroundVolume` schema doc) get a calmer baseline.

The slider also becomes useful in both directions: raise it for users who want louder, lower it for users who want quieter, instead of being a one-way attenuator.

## Area

Stream Deck plugin (defaults live in `@iracedeck/deck-core`; voice clips live in `@iracedeck/audio-assets`).

## Alternatives considered

- **Lower defaults only, keep the bake-in gain.** Doesn't fix the slider asymmetry — users who want louder voice still have a hot ceiling because the clips are already gain-staged hot.
- **Remove bake-in gain only, keep 100 % defaults.** First-run is still too loud even after de-juicing, because the voice channel sits at unity against game audio.
- **Re-render with a different filter philosophy (e.g., loudness-normalize to −16 LUFS).** Out of scope for this issue — could be a follow-up. `compare-loudness.mjs` already exists for that experiment.

### Files touched

- `packages/audio-assets/src/presets.mjs` — edit `RADIO_ENGINEER_FILTER` chain.
- `packages/deck-core/src/global-settings.ts` — three `.default(...)` updates plus their JSDoc blocks (the `backgroundVolume` doc explicitly mentions "loud relative to the engineer voice at unity"; the new-default narrative needs to mention 50/25/50 as the calmer baseline).
- `packages/audio-assets/voice/<voice>/**/*.mp3` — automatically regenerated when the filter hash changes; commit the regenerated processed-asset cache if it's tracked.

### Verification

- Re-build: `pnpm --filter @iracedeck/audio-assets build` (forces filter-chain reprocess on cache miss).
- Spot-check three clips before/after: confirm peaks sit closer to source level, no audible clipping, and the bandpass character is preserved.
- Fresh-install plugin in Stream Deck: confirm new defaults render in the Pit Crew PI volume controls.
- Existing users: their saved volume values are not changed (they only had to override because of the loudness — once the filter no longer bakes in gain, their existing saved values may sound too quiet, which is acceptable; the slider now works the way it should).

### Cross-cutting artifacts (per `.claude/rules/build-and-commit.md`)

- **Plugin manifests** — no change.
- **Plugin packages** — `iracing-plugin-stream-deck` and `iracing-plugin-mirabox` both consume `deck-core` defaults; both pick up the new values on rebuild. No code change needed.
- **Website** — `packages/website/src/content/docs/docs/actions/driving/pit-crew.md` if it documents specific default volume numbers.
- **Action documentation** (`docs/`) — same check.
- **Skills** — none.
- **Rules / CLAUDE.md** — `.claude/rules/global-settings.md` if it lists default values.

### Tests

- Unit test in `packages/deck-core` (or wherever `GlobalSettingsSchema` is exercised) asserting the three new defaults.
- Smoke-test the regenerated audio asset pipeline doesn't error on the simplified filter chain.

### Related

- **#471** — `backgroundVolume` was already lowered to 35 specifically because the ambience was overpowering the voice at unity. With voice no longer baked-in hot, the same intelligibility intent now wants 25.

niklam:
## Related Issue

Fixes #522

## What changed?

Two related volume fixes for the Race Engineer audio path:

1. **Radio filter no longer bakes in makeup gain.** `RADIO_ENGINEER_FILTER` in `packages/audio-assets/src/presets.mjs` drops the `volume=8dB` pre-gain and the cancelling `volume=-6dB,volume=6dB` pair. The bandpass + soft-clip + brick-wall limiter remain — soft-clip and limiter are now quiescent safety nets at unity input rather than catching peaks the makeup gain pushed past 0 dBFS. The cache hash changes, so every voice clip reprocesses on the next build automatically.

2. **Lowered user-facing volume defaults** in `packages/deck-core/src/global-settings.ts` so the slider has headroom in both directions instead of acting as a one-way attenuator:

   | Setting | Before | After |
   |---|---|---|
   | `raceEngineerVolume` | 100 | 50 |
   | `backgroundVolume` | 35 | 25 |
   | `radarVolume` | 100 | 50 |

Schema test, Pit Crew PI defaults, website docs, and internal docs updated to match. The runtime `VOLUME_MAX` fallback in `pit-crew.ts` is left alone (it's a fail-loud-on-bad-cache safety net, not a UX default), and the scenario-harness bootstrap defaults stay at 100 (dev tool, independent of user UX).

## How to test

1. Pull the branch and run `pnpm install && pnpm build`. Watch for `Audio assets: 113 processed, … (filter hash 0f645643)` — confirms the filter chain change invalidated the cache and reprocessed every voice clip.
2. Run `pnpm test` (3773 tests should pass).
3. Restart Stream Deck so the plugin reloads with the reprocessed voice clips.
4. Clear persisted globals if needed to see fresh-install defaults, then in the Pit Crew PI confirm the three sliders show **50 / 25 / 50** for Race Engineer / Background / Radar.
5. Hit each **Test** button — voice should sound noticeably less hot than before, with the bandpass radio character preserved; Background and Radar should sit naturally under the voice.
6. Existing users keep their saved values; only fresh installs see the new defaults.

## Checklist

- [x] New code has unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Lowered default Pit Crew audio levels: Race Engineer 100→50, Radar 100→50, Background 35→25.
  * Simplified Race Engineer audio processing chain to improve peak safety and remove baked-in makeup gain.

* **Documentation**
  * Updated Pit Crew docs and UI defaults to reflect the new audio defaults.


niklam:
@coderabbitai both suggestions applied in 3f650b88:

- **`docs/plugins/core/actions/pit-crew.md`** — added the missing `Background Volume` bullet (default 25), inserted between Race Engineer Volume and Radar Volume to match both the PI ordering and the section's "Race Engineer settings appear first" prose. Includes the #471 RE-off muting note for parity with the website doc.
- **`packages/deck-core/src/global-settings.test.ts`** — added `raceEngineerVolume: 50` to the `toMatchObject` payload so the regression coverage is symmetric across all three volume defaults.

PR #538, opened 2026-05-11
improve(release): use HEAD commit count for manifest build number (#536)

### LINKED ISSUE #536: Append commit-count build number to plugin manifest versions
## What would you like?

Populate the 4th component (`build`) of the `Version` field in both plugin manifests with the HEAD commit count (`git rev-list --count HEAD`) at release time, instead of always writing `.0`.

Example: a release tagged `v1.17.0-pre.1` cut when `git rev-list --count HEAD` returns `1542` would produce:
- `packages/iracing-plugin-stream-deck/com.iracedeck.sd.core.sdPlugin/manifest.json` → `"Version": "1.17.0.1542"`
- `packages/iracing-plugin-mirabox/com.iracedeck.sd.core.sdPlugin/manifest.json` → `"Version": "1.17.0.1542"`

The `package.json` semver remains unchanged (`1.17.0-pre.1` etc.) — only the manifest Version is extended.

## Use case

Today `scripts/release-hooks.mjs` strips any `-pre.N` / `+meta` suffix from the version and writes `{numericVersion}.0` to both plugin manifests. Elgato's manifest schema requires strict `{major}.{minor}.{patch}.{build}` 4-part numeric form, so semver pre-release labels can't survive in the manifest. The consequence: every pre-release of a given version (e.g. `1.17.0-pre.1`, `1.17.0-pre.2`, `1.17.0` final) ships with the same manifest version `1.17.0.0`, blocking any meaningful pre-release flow because Stream Deck installer comparison treats them as the same artifact.

Using `git rev-list --count HEAD` for the build slot makes every release a unique 4-part version, with the final naturally greater than its preceding pre-releases (because release-it commits the version bump between runs, advancing the count).

## Area

Stream Deck plugin

## Alternatives considered

- **Committed counter file** — explicit and meaningful (release iteration N), but adds state to maintain and a small merge-conflict surface area.
- **Date-based (YYMMDD + sub-counter)** — encodes when the build was made, but needs same-day collision handling.
- **\`\$GITHUB_RUN_NUMBER\`** — monotonic and stateless, but only works inside Actions and doesn't survive workflow recreation.
- **Pinning to \`origin/master\`** instead of \`HEAD\` — rejected because it breaks test releases from feature branches (two test releases on the same branch would collide since master's count doesn't change between them).

`HEAD` commit count wins for: zero state, naturally monotonic per branch, every release-it run guaranteed to bump the count (release commit adds 1), supports test releases from feature branches.

## Additional context

Single-file change in \`scripts/release-hooks.mjs\`:
- Add \`execFileSync(\"git\", [\"rev-list\", \"--count\", \"HEAD\"], …)\` lookup
- Replace \`\${numericVersion}.0\` with \`\${numericVersion}.\${buildNumber}\` for both manifests
- Update dry-run output to show the would-be build number
- Refresh the block comment at lines 39–43 that currently says \"every 1.15.0 pre-release ships with the same manifest Version '1.15.0.0'\"

Out of scope: any rename/restructure of \`release-hooks.mjs\`, changes to release-it commit/tag format, or mitigation for the rare edge case of two unmerged branches at the same commit count cutting test releases simultaneously.

niklam:
## Related Issue

Fixes #536

## What changed?

`scripts/release-hooks.mjs` now populates the 4th component of the plugin manifest `Version` field from `git rev-list --count HEAD` instead of hardcoding `.0`. Every release-it run (pre-release or final) produces a unique 4-part manifest version, so pre-releases of the same semver (e.g. `1.17.0-pre.1`, `1.17.0-pre.2`, `1.17.0` final) no longer collide on the Stream Deck side.

Monotonicity falls out naturally: release-it commits the version bump between runs, which advances the commit count by at least one, so the final always outranks its preceding pre-releases. Test releases cut from feature branches each get their own count progression (HEAD-based, not pinned to `origin/master`) so different branches don't share build numbers.

Block comment in the script refreshed to reflect the new behavior; dry-run preview now shows the actual build number it would write.

## How to test

```bash
# Direct hook dry-run (no side effects):
RELEASE_IT_DRY_RUN=1 node scripts/release-hooks.mjs 1.17.0-pre.1
# Expect the manifest preview line to read "version 1.17.0.<N>"
# where <N> matches `git rev-list --count HEAD`.

# Full release-it dry-run:
pnpm release -- minor --preRelease=pre --dry-run
```

## Checklist

- [ ] New code has unit tests — N/A, no existing test coverage for `scripts/release-hooks.mjs`; the change is verified via direct hook dry-run.
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — both manifests share the same hook; no per-plugin changes needed.

## Summary by CodeRabbit

* **Chores**
  * Version numbering has been updated to include unique build identifiers alongside standard version components, ensuring every release and pre-release build receives a distinct version string.

[Review Change Stack](https://app.coderabbit.ai/change-stack/niklam/iracedeck/pull/538)

### DIFF
--- a/scripts/release-hooks.mjs
+++ b/scripts/release-hooks.mjs
@@ -39,12 +39,22 @@ const packageJsonPaths = readdirSync(packagesDir, { withFileTypes: true })
 // Bump Version in manifest.json files. Elgato's manifest schema requires a
 // strict 4-part numeric format `{major}.{minor}.{patch}.{build}`
 // (^(0|[1-9]\d*)(\.(0|[1-9]\d*)){3}$), so semver pre-release / build metadata
-// suffixes must be stripped. Every 1.15.0 pre-release and the final 1.15.0
-// therefore ship with the same manifest Version "1.15.0.0".
+// suffixes must be stripped. The build slot is populated from
+// `git rev-list --count HEAD` so each release (pre or final alike) gets a
+// unique 4-part version; the final naturally outranks its preceding
+// pre-releases because release-it commits the version bump between runs,
+// which advances the commit count.
 const manifestPaths = [
   "packages/iracing-plugin-stream-deck/com.iracedeck.sd.core.sdPlugin/manifest.json",
   "packages/iracing-plugin-mirabox/com.iracedeck.sd.core.sdPlugin/manifest.json",
 ];
+
+const numericVersion = version.replace(/[-+].*$/, "");
+const buildNumber = execFileSync("git", ["rev-list", "--count", "HEAD"], {
+  cwd: root,
+  encoding: "utf-8",
+}).trim();
+const manifestVersion = `${numericVersion}.${buildNumber}`;
 
 // release-it runs before:bump hooks even in dry-run mode, which would otherwise
 // modify real package.json / manifest.json files and stage them with `git add`.
@@ -52,9 +62,7 @@ const manifestPaths = [
 if (process.env.RELEASE_IT_DRY_RUN === "1") {
   console.log(`  [dry-run] Would bump ${packageJsonPaths.length} package.json files to version ${version}:`);
   for (const rel of packageJsonPaths) console.log(`    - ${rel}`);
-  console.log(
-    `  [dry-run] Would bump ${manifestPaths.length} manifest.json files to version ${version.replace(/[-+].*$/, "")}.0:`,
-  );
+  console.log(`  [dry-run] Would bump ${manifestPaths.length} manifest.json files to version ${manifestVersion}:`);
   for (const rel of manifestPaths) console.log(`    - ${rel}`);
   process.exit(0);
 }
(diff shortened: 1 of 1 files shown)


niklam:
Thanks for the review @coderabbitai. Both findings declined, reasoning below.

**`/* eslint-env node */` at top of file (outside-diff, Major)**

The `pnpm lint` script globs `packages/*/src/**` — `scripts/` is intentionally out of scope, and CI's `ESLint check` does not touch this file (confirmed against `eslint.config.js` and `package.json`'s `lint`/`lint:fix` scripts). Peer Node scripts in this directory (`release.mjs`, `generate-icon-previews.mjs`, `generate-icon-defaults.mjs`, etc.) do not carry an `eslint-env` directive either. Adding it to `release-hooks.mjs` alone would be a one-off inconsistency. If we ever want `scripts/` linted, the right fix is to extend `eslint.config.js` so `**/*.mjs` picks up `globals.node` once for every script — separate scope.

**Validate `numericVersion` and `buildNumber` before composing `manifestVersion` (nitpick)**

`version` comes from release-it which always passes a valid semver string; if it didn't, the next step (`npm version`) would reject it anyway. `git rev-list --count HEAD` either returns a positive integer or exits non-zero, which `execFileSync` already throws on. Both failure modes are covered. The rest of the script doesn't validate its inputs (no schema check on the parsed package.json files, no shape check on `manifest.json`), so adding defensive regex only at this spot would be inconsistent with the rest of the file's style.

PR #723, opened 2026-06-27
fix(pi): open external Property Inspector links in default browser (#243)

### LINKED ISSUE #243: fix(pi): open Downloads and Documentation links in default browser
## What would you like?

External links in Property Inspectors (Downloads link in `version.ejs`, Documentation link in `docs-link.ejs`) currently open in the Stream Deck app's built-in browser. They should open in the user's default browser instead.

### Solution

Use `streamDeck.system.openUrl(url)` from the Elgato SDK, which opens URLs in the default browser. Since this API is only available in the plugin context (not the PI), the PI must send the URL to the plugin via `sendToPlugin`, and the plugin must handle it by calling `streamDeck.system.openUrl()`.

Reference: https://docs.elgato.com/streamdeck/sdk/guides/system

### Affected partials

- `packages/stream-deck-plugin/src/pi-templates/partials/version.ejs` — Downloads link
- `packages/stream-deck-plugin/src/pi-templates/partials/docs-link.ejs` — Documentation link

### Implementation notes

- PI side: intercept link clicks, prevent default, send URL to plugin via `sendToPlugin`
- Plugin side: listen for the message and call `streamDeck.system.openUrl(url)`
- Custom URL schemes (e.g., `my-app://`) are not supported by the SDK
- Both `stream-deck-plugin` and `mirabox-plugin` PI templates need updating

niklam:
## Related Issue

Fixes #243

## What changed?

Links in a Property Inspector (Documentation in `docs-link.ejs`, Downloads in `version.ejs`, and the "View full list of template variables!" links in the Chat and Telemetry Display actions) opened in the deck app's small built-in browser. No PI link ever wants that, so a single delegated click handler — `installExternalLinkHandler()` in the new `packages/pi-components/src/components/external-links.ts`, run on load from the `pi-components.js` bundle entry — now reroutes **every** external `http(s)` anchor click through sdpi-components' built-in `openUrl` event.

That event reaches the OS default browser on every host with no plugin-side code: Elgato handles it natively, the VSD (Mirabox) host (which speaks the Elgato PI protocol directly) receives it, and the Ulanzi PI bridge already translates `openUrl` → its `openurl` cmd. The handler resolves the anchor's normalized `.href` and only reroutes `http(s)` (the SDK can't open other schemes, e.g. `mailto:`/`app://`); it only intercepts when sdpi-components is loaded, so a load failure falls back to the link's default behavior; the send is fire-and-forget (rejections swallowed) and the listener installs once per document.

The `#4a90d9` PI link accent was also consolidated into a `--ird-link-color` CSS variable in `head-common.ejs`, replacing the per-link literals.

## How to test

1. Build and load the plugin (`pnpm build`), open any action's Property Inspector.
2. Click the **Documentation** link (bottom of most actions) and the **Downloads** link in the version footer — both should open in your real default browser, not an in-app window.
3. Open the **Chat** or **Telemetry Display** action and click **"View full list of template variables!"** — same behavior.
4. Best-effort: repeat on Mirabox / Ulanzi hardware if available (see note below).

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — shared `@iracedeck/pi-components` bundle, consumed by all three plugins; no per-plugin wiring needed

## Note on host behavior

`openUrl` is best-effort on non-Elgato hosts: the PI sends the event, but the Stream Dock (Mirabox) / UlanziStudio host must honor it. The SDK gives no success callback, so the handler can't conditionally fall back — same caveat the existing version-check changelog-opener carries. Elgato (the primary target) is reliable. A follow-up issue tracks verifying/​hardening this on Mirabox/Ulanzi.

## Summary by CodeRabbit

* **Bug Fixes**
  * Links in action settings and related PI screens now open in the user’s default web browser instead of the deck app’s built-in browser.
  * External `http`/`https` links are handled more reliably, including clicks on elements nested inside a link.
* **Documentation**
  * Updated PI template guidance and changelog notes with clearer cross-references and an explanation of the default-browser external link behavior.
* **Style**
  * Standardized the PI link/accent color using a shared styling variable for consistent focus and version link accents.

### DIFF
--- a/.claude/rules/pi-templates.md
+++ b/.claude/rules/pi-templates.md
@@ -116,8 +116,8 @@ Located in `packages/pi-components/partials/`:
 - **global-graphic-defaults.ejs** - Global graphic scale default (50-150%, default 100%) in accordion
 - **global-flag-flash.ejs** - Global flag-flash duration default (0-30 seconds, default 15, step 1; `0` = flash forever) in accordion. See issue #490.
 - **global-common-settings.ejs** - Global common settings (window focus, SimHub server) in accordion
-- **docs-link.ejs** - Documentation link to the action's page on iracedeck.com (conditional, hidden when no URL mapped)
-- **version.ejs** - Version footer with downloads link
+- **docs-link.ejs** - Documentation link to the action's page on iracedeck.com (conditional, hidden when no URL mapped). Opens in the default browser (see _External Links_ below).
+- **version.ejs** - Version footer with downloads link. Opens in the default browser (see _External Links_ below).
 
 ## Shared CSS Classes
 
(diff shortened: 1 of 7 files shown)


niklam:
Out-of-scope items surfaced by review are tracked as follow-ups: #724 (verify/harden PI `openUrl` on Mirabox/Ulanzi hosts) and #725 (replace inline link-color styles with `--ird-link-color`).

PR #729, opened 2026-06-27
improve(fuel-dial): match Add Amount readout to telemetry and fix Fuel Target toggle-off re-arm (#726)

### LINKED ISSUE #726: improve(fuel-dial): match displayed/requested fuel to what iRacing actually applies
**Target:** `release/2.0` (2.0.0 milestone). Fuel Dial is a 2.0 action.

## What would you like?

Make the Fuel Dial's requested fuel match what iRacing actually applies, and show that authoritative value immediately instead of optimistically guessing the dialed amount and re-syncing it a couple of seconds later. Two coupled symptoms from user feedback (Discord, @LT4650):

1. **Unit/rounding discrepancy.** With display units = gallons, dialing "+1" does not put 1 gallon in the tank — the black box ends up adding ~0.8 gal. iRacing's fuel cell works internally in liters and `pit.fuel` only accepts whole liters, so "+1 gal" (≈3.79 L) is rounded and the car reports back a different gallon figure. Even the chat macro `#fuel +1 g` shows the same ~0.8 gal result (reproduced on the BMW M2).
2. **Display lag / "guess then settle".** On rotate the readout immediately bumps by the dialed step (e.g. +1), then ~2–3 s later snaps to the value the black box actually holds. That is the optimistic local value being overwritten when telemetry re-seeds.

## Use case

The dialed number, the touch-strip readout, and the in-sim black box should always agree. Today a driver dials "+1 gal", sees "+1" for a couple of seconds, then watches it jump to ~"+0.8" — confusing mid-stop, and worse in VR where a glance has to be trusted. The value should be correct and stable on the first render.

## Area

Action (existing) — Fuel Dial (`com.iracedeck.sd.core.fuel-dial`).

## Current behavior (release/2.0)

- All fuel comms go through the iRacing API: `pit.fuel(liters)` / `pit.clearFuel()` in `packages/iracing-actions/src/actions/fuel-dial/fuel-dial.ts`. `pit.fuel` takes **whole liters** (the broadcast value is an unsigned int; `pit.fuel(0)` means "keep existing").
- The dial keeps a local `dialValueLtr` and renders it optimistically on every rotate (`onDialRotate` → `scheduleSend`/`render`).
- Telemetry re-seeds `dialValueLtr` from `PitSvFuel` (the add iRacing reports back), but only after `USER_ACTIVITY_GRACE_MS` (3000 ms) with no rotation — this grace window is the visible "settle a couple seconds later" delay.
- Display-unit conversion (`fuelFromDisplayUnits` / `fuelToDisplayUnits`) plus the whole-liter broadcast is where the gallon discrepancy enters.

By contrast the **Fuel Service** action reads `PitSvFuel` straight from telemetry and shows the black box value essentially immediately (one ~16.7 ms tick at 60 Hz), with no local guess to reconcile.

## Proposed directions (to be evaluated — not mutually exclusive)

- **A. Closed-loop display (send → read back).** After broadcasting `pit.fuel`, drive the displayed add from `PitSvFuel` (the authoritative black box value) rather than the local guess, accepting ~one telemetry tick of latency — what Fuel Service effectively does. Removes the resync jump. Open question: keep the dial responsive during a fast spin (e.g. show the in-flight target during the throttle window, then settle to `PitSvFuel`).
- **B. Calibrate the effective per-step delta.** Learn how much a request actually moves the tank by watching `PitSvFuel` deltas: if "+1 gal" yields +0.8 gal, future steps and the readout reflect 0.8/step → no discrepancy. Also covers the "a 1 L request on a gallons car does nothing, so try 2 L, 3 L… until the value actually moves" case.
- **C. Shorten/replace the 3 s re-seed grace** so the authoritative value appears promptly rather than after a multi-second window (likely a consequence of A).

## Alternatives considered

- Leave the optimistic guess and just shorten the grace window — fixes only the lag, not the unit discrepancy.
- Round the dialed value to whole gallons before sending — doesn't help, because the car's internal liter rounding (not our conversion) is the source.

## Additional context

- Reported by @LT4650; analysis in the Discord thread (BMW M2 repro). Reporter settings: Add Amount mode, Amount 1, Unit Auto (from iRacing → gallons), Press = Toggle Fueling.

## Affected artifacts (update in the same PR)

- [ ] `packages/iracing-actions/src/actions/fuel-dial/fuel-dial.ts` + `fuel-dial.test.ts`; possibly `packages/deck-core/src/fuel-telemetry.ts` (+ its test)
- [ ] Action docs: `docs/plugins/core/actions/fuel-dial.md`
- [ ] Website action page: `packages/website/src/content/docs/docs/actions/pit-service/fuel-dial.md`
- [ ] Changelog (`packages/website/src/content/docs/changelog.mdx`) on merge to `release/2.0`
- [ ] Rules/skills if any behavior convention changes

niklam:
## Related Issue

Fixes #726

## What changed?

Two related Fuel Dial fixes for the unreleased 2.0 action:

- **Add Amount display follows telemetry (#726).** The readout and bar now show the live pit request iRacing reports (`PitSvFuel`) — what the game actually banked after its whole-litre rounding — instead of the optimistically-dialed value that resynced a beat later (the "+1 → +0.8 gal" jump). The dialed value still drives what is sent (`pit.fuel`); only the display is decoupled onto telemetry, clamped to `[0, capacity]`.
- **Fuel Target toggle-off no longer re-arms.** After a toggle-off cleared the request, the continuous fill-to monitor re-broadcast `pit.fuel` during iRacing's telemetry-lag window (while fuel-fill still read ON), silently re-arming the fueling the user just turned off. It is now gated on `!lastPitWasClear`, so a deliberate clear sticks and continuous re-sends resume only once fueling is genuinely re-armed.

Action docs (`docs/plugins/core/actions/fuel-dial.md`), the website action page, and the in-development 2.0 changelog line are updated to match.

## How to test

In iRacing with a Stream Deck+ Fuel Dial:

- **Add Amount:** set Units to Gallons and dial a fuel amount. The readout should immediately show the amount iRacing actually applies (no "+1" that jumps to "+0.8" a couple seconds later).
- **Fuel Target:** with a target set and fueling on, press to toggle fueling off. It should stay off — it must not silently re-arm a moment later.

## Checklist

- [x] New code has unit tests
- [ ] Changes registered in all applicable plugins (Stream Deck, Mirabox) — N/A: shared action logic in `@iracedeck/iracing-actions`; no plugin registration or manifest change

## Summary by CodeRabbit

* **Bug Fixes**
  * Fuel Dial now keeps the displayed **Add Amount** aligned with what iRacing actually banks, so the readout no longer briefly shows a dialed estimate before correcting itself.
  * Continuous fueling behavior is more reliable during short telemetry delays, preventing unintended re-sends and reducing display flicker or stale values.

* **Documentation**
  * Updated Fuel Dial guidance to better explain manual, target, and autofuel modes and how the on-screen fuel readout behaves.


niklam:
**Re: CodeRabbit pre-merge "Docstring Coverage" warning (50% < 80%)** — not acting on this, intentionally.

The coverage figure is driven by the many `it(...)` test callbacks in the diff, which are documented by their `it("…")` descriptions per this repo's testing conventions (`.claude/rules/testing.md`), not by JSDoc. The one new non-test helper, `getTelemetryCallback`, already has a docstring, and the production change is to `displayAddLtr` — an already-documented method whose JSDoc was updated in this PR. There are no new undocumented production functions, and the project has no 80%-docstring policy, so adding JSDoc to test callbacks purely to satisfy the threshold would be noise against the codebase's style.

CI (lint/format/test) is green and CodeRabbit approved with no actionable comments, so there's nothing further to change here.

Issue #710, opened 2026-06-27
Telemetry Display: race_ahead/race_behind can resolve to the wrong car on duplicate-rank position collisions


niklam:
### What happens

`race_ahead` / `race_behind` (in `buildTemplateContextFromData`, via `findDriverByRacePosition`) select the neighbouring car from the pit-road-blended `resolveRacePositions` array. That blend keeps an on-track car at its live lap-order rank but snaps a pit-road / retired / frozen-rank car to its official `CarIdxPosition`. The two sources can collide, so the resolved array can contain **two carIdx entries with the same overall position**. `findDriverByRacePosition` then returns whichever duplicate appears first in the `DriverInfo.Drivers` array — which may be the wrong car.

### Example

An on-track car at live-order P4 and a pitting car at official P4 both end up as "P4" in the resolved array. A player at P3 looking one position back (`race_behind`) can resolve to whichever of the two is listed first, not the car actually adjacent in the standings.

### Notes

- This is **pre-existing**: `resolveRacePositions` already merged the calculated order with official `CarIdxPosition` before the #700 work. The live-order injection from #700 just makes the divergence (and therefore collisions) more frequent for pit/finished cars, which is why it surfaced now.
- Surfaced by the #700 code review (PR #709), deliberately left out of that PR's scope because a correct tie-break is non-trivial (which car is "really" at P4 when two share the rank?).

### Possible direction

Make the standings used for *neighbour selection* collision-free — e.g. resolve ties deterministically by car-on-track order or by official position, or select `race_ahead`/`race_behind` from a de-duplicated ranking rather than the pit-blended array.

### Affected artifacts

- `packages/iracing-sdk/src/template-context.ts` (`resolveRacePositions`, `findDriverByRacePosition`)
- Tests in `packages/iracing-sdk/src/template-context.test.ts`

Issue #697, opened 2026-06-23
Race Engineer: position callouts read too optimistic (frozen running-order strands cars that go NotInWorld)


niklam:
## What happened?

The Race Engineer announces a race position that is **better (too optimistic)** than the driver's actual position. Multiple users have reported it; for at least one driver the spoken position was wrong by **several places**. In a captured case the engineer said "We're currently P7" while the driver was genuinely P8.

The spoken position is derived from iRaceDeck's self-managed "frozen" running order (`getLivePosition()` → `calculateFrozenRacePositions()`, introduced in #603), **not** iRacing's authoritative `PlayerCarPosition`. That frozen order can rank the player ahead of cars that are genuinely still ahead, so the readout reads optimistic. The more cars stranded at once, the more positions optimistic the call.

## Expected behavior

The spoken position should match iRacing's official classification (`PlayerCarPosition` / `CarIdxPosition`, or the class equivalent). If the driver is P8, the engineer says P8.

## Steps to reproduce

1. Race (single- or multi-class) where a car **ahead** of you briefly goes `NotInWorld` — a connection blip, a tow, or any out-of-world moment longer than ~1 second — and then returns to racing.
2. Continue racing past where that car was last seen on track.
3. Trigger any position readout (overtake gain/loss, lap-completed position change, or the every-3-laps race-status cadence).
4. The engineer announces a position one or more places better than iRacing's official position.

Reproduced from captured telemetry — see Root cause / Evidence.

## Root cause

`getLivePosition()` (`packages/sim-events-iracing/src/translator.ts`) builds its own order via `calculateFrozenRacePositions()` (`packages/sim-events-iracing/src/diff/race-finish.ts`) and never falls back to iRacing's authoritative overall position. The frozen model (#603) deliberately "freezes" a car at its last on-track score whenever it goes `NotInWorld` (blip / tow / finisher leaving), so the player only passes it by racing past that frozen point. The blind spot:

- `updatePositionTracking()` only **unfreezes** a returning car when its live score comes back within `TELEPORT_THRESHOLD` (0.05 lap) of its stale anchor. A car out of world long enough to advance more than that (≈1.2 s of racing) returns far from its anchor, is misclassified as a permanent "teleport," and **stays frozen at the stale score** — the anchor never rolls forward.
- A car ahead that was last anchored near/behind the player's track position then sits stranded **below** the player in the frozen order. The player "passes" the stale anchor and is ranked one (or more) places too high, disagreeing with iRacing.
[...]

PR #734, opened 2026-06-29
fix(sim-events-iracing): keep towed cars at their position until they rejoin the track (#733)

### LINKED ISSUE #733: Towed cars jump ahead in the running order while sitting in the pits
## Summary

A car that is towed into the pits is briefly ranked **ahead** of cars it shouldn't be, then drops back once you drive past the pit area. It shows up in the Race Engineer's spoken position, the Session Info → Position readout, and the Telemetry Display / Chat / Race Admin driver-info prefixes. Reported in-sim: "I lost a position to someone who towed, then gained it back when I passed the pits," repeatedly.

## Cause

The running order is the self-managed "frozen positions" calculation in `packages/sim-events-iracing/src/diff/race-finish.ts`, which is the single source behind the Race Engineer, Session Info Position, and the template prefixes. A towed car is frozen at its last on-track score, but #697 (shipped in v1.22.2) changed the release rule to "release the moment it resumes any forward motion." A car towed into its pit box sits at a high pit-box `CarIdxLapDistPct`, so the instant it creeps forward during service it released and snapped its rank to that pit-box score — ahead of cars still racing earlier on the same lap. #710 then removed the template's pit-road position blend, exposing the same jump on the Telemetry Display. This is a regression of the behavior released in v1.22.2.

## Expected

A towed car holds its on-track position for the whole pit stop and only drops back once it rejoins the track — it never momentarily jumps ahead of cars still racing.

## Fix

Hold a frozen (towed) car at its on-track anchor while it is on pit road and release it only once it has rejoined the racing surface and is moving. The "on pit road" test must use `CarIdxTrackSurface` (`InPitStall` / `AproachingPits`), not `CarIdxOnPitRoad`, which real telemetry shows is unreliable (it reads `true` for cars on track and `false` for cars sitting in their box). Normally-pitting cars (driving in under their own power) are not frozen and are unaffected.

## Affected artifacts

- `@iracedeck/sim-events-iracing` — the canonical frozen order (`race-finish.ts`) and its tests
- Race Engineer position callouts, Session Info → Position, Telemetry Display / Chat / Race Admin template prefixes (all consume the same order; no per-consumer change needed)
- Changelog (`changelog.mdx`)

niklam:
## Related Issue

Fixes #733

## What changed?

A car towed into its pit box sits at a high pit-box `CarIdxLapDistPct`. The #697 release-on-first-motion rule released the frozen car the instant it crept forward during service, snapping its rank to that pit-box score and jumping it ahead of cars still racing earlier on the same lap; #710 then removed the template's pit-road blend that had masked this on the Telemetry Display.

This holds a frozen (towed) car at its on-track anchor while it is on pit road and releases it only once it has rejoined the racing surface and is moving. The "on pit road" test uses `CarIdxTrackSurface` (`InPitStall` / `AproachingPits`) — **not** `CarIdxOnPitRoad`, which real telemetry shows is unreliable (it reads `true` for cars on track and `false` for cars sitting in their box). `race-finish.ts` is the single source behind the Race Engineer position callouts, Session Info → Position, and the Telemetry Display / Chat / Race Admin prefixes, so all three are fixed from one place. Normally-pitting cars (driving in under their own power) are never frozen and are unaffected.

It also removes the phantom "lose then re-gain": the towed car no longer jumps ahead and back, so there is no spurious gain for the overtake retirement classifier to handle.

## How to test

1. In a race, have a rival get towed (or tow yourself). On the Telemetry Display, Session Info → Position, and the Race Engineer callouts, confirm the towed car holds its position while in the pits and does **not** momentarily jump ahead of cars still racing.
2. Confirm it drops back to its real (dropped-back) position only once it drives back out onto the track.
3. Confirm a car that drives into the pits normally still shows its live position (it is not frozen).

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — N/A: the fix is in the shared `@iracedeck/sim-events-iracing` running order consumed by every plugin; no plugin registration, manifest, or PI change

## Summary by CodeRabbit

* **Bug Fixes**
  * Cars towed into the pits no longer temporarily jump ahead in the running order while in their pit box.
  * Race Engineer spoken updates and position readouts now correctly reflect the car’s place during pit stops.
  * Towed/teleported cars remain “frozen” until they are both moving continuously and back off pit road; they only resume normal running-order placement once rejoining the racing surface.

### DIFF
--- a/packages/sim-events-iracing/src/diff/overtakes.test.ts
+++ b/packages/sim-events-iracing/src/diff/overtakes.test.ts
@@ -1223,12 +1223,13 @@ describe("diffOvertakes — finished / retired cars (#603)", () => {
     expect(state.positionFrozen.has(1)).toBe(true);
     expect(state.pendingOvertakePos).toBe(-1);
 
-    // Rival starts rolling again → released this tick, drops to its live back-of-
-    // field score, and the player inherits P2. The gain must read fromRetirement
-    // (the rival fell back; the player didn't pass it on track), NOT a real pass.
+    // Rival drives back out of the pit onto the track → released this tick (moving
+    // AND off pit road), drops to its live back-of-field score, and the player
+    // inherits P2. The gain must read fromRetirement (the rival fell back; the
+    // player didn't pass it on track), NOT a real pass.
     tickAt(
       state,
-      mkTel([0.51, 0.06, 0.81], { trackSurface: [TrkLoc.OnTrack, TrkLoc.InPitStall, TrkLoc.OnTrack] }),
+      mkTel([0.51, 0.06, 0.81], { trackSurface: [TrkLoc.OnTrack, TrkLoc.OnTrack, TrkLoc.OnTrack] }),
       200,
       emit,
     );
@@ -1240,7 +1241,7 @@ describe("diffOvertakes — finished / retired cars (#603)", () => {
     // Hold elapsed → emits with fromRetirement=true (no "Nice pass").
     tickAt(
       state,
-      mkTel([0.512, 0.06, 0.812], { trackSurface: [TrkLoc.OnTrack, TrkLoc.InPitStall, TrkLoc.OnTrack] }),
+      mkTel([0.512, 0.06, 0.812], { trackSurface: [TrkLoc.OnTrack, TrkLoc.OnTrack, TrkLoc.OnTrack] }),
       200 + OVERTAKE_HOLD_MS,
       emit,
     );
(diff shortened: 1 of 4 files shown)


niklam:
Addressed the `AproachingPits` nitpick in 5044fe3d — added a regression case that holds a towed car frozen while it's on the pit lane (`AproachingPits`), not just in the box (`InPitStall`), and releases it only once it's back on track. That locks down the second half of the pit-road gate's contract.

Issue #732, opened 2026-06-29
Add a dynamic Race Admin / RC car selector (profile-based)


niklam:
### What would you like?

A dynamic Race Admin / Race Control workflow that auto-populates one button per car in the current iRacing session, so admins and race control no longer have to hand-build a page for every possible car number ahead of time. Pressing a car selects it as the admin target, and the deck navigates to a set of Race Admin / RC commands (black flag, clear penalty, drive-through, stop-and-go, EOL, DQ, wave-around, etc.) that act on that car.

The proposed implementation is **profile-based**, using Elgato Stream Deck profiles shipped with the plugin:

- **Race Admin car selector profile** ("Available Cars") — auto-populated with one button per car number in the session, using placeholder buttons filled from live session data (CarIdx → car number / name / class). Spans multiple pages so even 40–60-car enduro fields fit.
- **Per-car Race Admin command profile** — the admin/RC command buttons. The target car is chosen dynamically by whichever selector button was pressed; the selected car is stored globally so every admin button acts on that target.
- **Navigation** — Main → Race Admin (Available Cars) → Car #123 → admin commands, with **back buttons** at each level to return to the selector and to the main profile.
- The **Main profile must also ship with the plugin**, because Stream Deck plugins can only switch to profiles distributed with the plugin — so we need a known profile to return to. See https://docs.elgato.com/streamdeck/sdk/guides/profiles/.
- Optional related profile: a **Pit Service profile** that appears automatically when the car enters pit road.

### Use case

Hosted sessions, leagues, and especially endurance races can have 40–60 cars, with car numbers that change every session and sometimes use 3-digit numbers. Today admins build a separate page (or several) for every possible number — e.g. one user maintains 3+ pages covering 00–99 and would need many more to cover 3-digit numbers. This is a large, repetitive, per-session setup cost. Auto-populating cars from the live session removes that setup entirely and makes per-car race control practical on the deck.

### Area

Stream Deck plugin (introduces new actions; relies on Elgato profile switching).

### Alternatives considered

**Standalone placeholder buttons (no profile switching)** — "placeholder" command buttons you drop on any existing page that fill themselves with a car number once the session starts and are selectable by CarIdx. A working community proof of concept of this exists already (Caridx). It's simpler, but doesn't scale to large fields the way paged profiles do and doesn't provide the Main → selector → per-car-commands navigation structure. The profile approach is preferred here; the placeholder fill mechanism is reused inside the auto-populated selector profile.

### Additional context

[...]

PR #751, opened 2026-07-03
fix(race-admin): refuse admin commands aimed at non-user or own-car targets (#747)

### LINKED ISSUE #747: Race Admin: user-management commands (revoke admin) aimed at an AI car self-apply — revoked the host's own admin and ended the session
## User report

> May have found a bug though. I pressed "revoke admin" on another car (was AI race, that car wasn't admin anyway). It revoked my admin and shut the session immediately

Feature context: the #732 Race Admin car selector / Selected Car target (shipped to master in #746, unreleased — 1.24.0).

## Investigation findings

**The plugin's command was well-formed.** All driver-target paths (`selected-car`, `viewed-car`, `specific`) resolve to a digit-cleaned car number or `null`; a `null` target aborts in `buildAdminCommand` (`race-admin-commands.ts`), so a key press can never send a bare `!nadmin`. The dispatched string was `!nadmin #<carnum>`, which matches iRacing's documented syntax (`!nadmin <driver>` where `<driver>` is `name.name` or `#<car number>` — [official docs](https://support.iracing.com/support/solutions/articles/31000133518-session-admin-chat-commands), [EdRacing reference](http://www.edracing.com/edr/iRacing-Admin.php)).

**The root cause is the target class, not the syntax.** `!admin` / `!nadmin` (and the `!chat` / `!nchat` driver variants) manage *users* — but an AI car has no connected user behind it. iRacing's behavior when the `#<num>` argument can't be matched to a user is undocumented, and the report indicates its failure mode applies the command to the *issuing* user instead: the sender's own admin was revoked, and since the host was the sole admin of an AI session, the session ended immediately. The selector and the car-targeted race-control commands (black flag, EOL, DQ, clear, wave-around) legitimately include AI cars — race-controlling AI is a primary use case — so the gap is specifically that **user-management modes accept car targets that cannot be users**.

`CarIsAI` is available per driver in `sessionInfo.DriverInfo.Drivers[]` (already consumed elsewhere, e.g. `sim-events-iracing/src/start-lights.ts`), so the target class is detectable at dispatch time.

## Proposed fix

1. Add a `targetsUser` flag to `RaceAdminModeMeta` for the modes whose `<driver>` must be a human user: `grant-admin`, `revoke-admin`, `enable-chat-driver`, `disable-chat-driver` (and decide `remove-driver` — removing an AI car may be legitimate; verify in-sim).
2. In `executeMode`, when a `targetsUser` mode resolves its target to a car whose `DriverInfo` entry has `CarIsAI === 1`, **do not send** — log a warning and give key feedback (e.g. `showAlert`) instead.
3. Consider the same guard for a target number that doesn't exist in the session at all (typo in `specific`, stale selection edge) — if iRacing's unmatched-argument fallback self-applies, any unresolvable target is dangerous for user-management commands.

## Verification needed (in-sim, AI session + hosted session)

- Confirm `!nadmin #<AI car>` self-applies (reproduces the report), and whether `!admin #<AI car>` does the same.
- Confirm `!remove #<AI car>` behavior (may validly remove the AI car).
- Confirm what `!nadmin #<nonexistent number>` does in a hosted session (typo case).
- Confirm race-control commands (`!black`, `!eol`, `!dq`, `!waveby`, `!clear`) still behave correctly against AI cars (expected: they work — no guard for these).

## Affected artifacts when implemented

- `packages/iracing-actions/src/actions/race-admin/` — mode metadata (`targetsUser`), dispatch guard in `race-admin.ts`, tests
- **Website** (`@iracedeck/website`) — Race Admin action page (document that user-management modes refuse AI targets); changelog: fold into the existing 1.24.0 car-selector feature line if fixed before release (per `.claude/rules/changelog.md`), else a Bug Fixes entry
- **Skills** (`iracedeck-actions`) — note the AI-target guard on the affected modes if mode descriptions change
- **Rules** (`.claude/rules/profiles-and-devices.md`) — note the user-vs-car target distinction in the #732 section if conventions change

niklam:
## Related Issue

Fixes #747

## What changed?

Guards Race Admin's driver-management commands against targets iRacing can't resolve to a real user — the failure mode observed in #747 is that iRacing applies such commands to the **sender**, so revoking admin on an AI car revoked the host's own admin and ended the session.

- **`targetsUser` mode flag** on the five user-management modes (Grant Admin, Revoke Admin, Enable Chat (Driver), Mute Driver, Remove Driver). At dispatch, the resolved target is classified via the new `classifyCarNumberTarget()` (`iracing-sdk` session-utils: `user` / `ai` incl. pace car / `unknown`, exact leading-zero-preserving number match) and the command is **refused** unless it classifies as a user — covering both the AI-car case from the report and the typo/not-in-session case.
- **`refusesSelfTarget` flag** on Revoke Admin: additionally refuses the sender's **own** car (new `getPlayerCarNumberFromSessionInfo()` via `DriverInfo.DriverCarIdx`) — revoking your own admin can end the session you're hosting, an easy slip with the viewed-car / selected-car targets. Grant Admin on yourself stays allowed.
- **Key feedback**: refusal flashes the host warning indicator via a new *optional* `IDeckActionContext.showAlert?()` — implemented by the Elgato adapter (keys only); other adapters omit it, so no interface churn there — plus a warn log with the reason.
- **Unchanged**: car-targeted race-control commands (`!black`, `!eol`, `!dq`, `!clear`, `!waveby`) still work against AI cars; the Type in Chat target is manual and bypasses the guards.
- Docs: changelog Bug Fixes entry (the hazard predates the #732 selector — viewed-car/specific targets can hit AI cars in released builds), website Race Admin page section, mode-meta JSDoc, profiles rule note, deck-core CLAUDE.md.

Scope decisions (revertible one-liners if in-sim verification disagrees): `remove-driver` is guarded as user-targeted pending verification of `!remove #<AI>`; unknown/typo numbers are refused for user-management modes only.

## How to test

1. Build and install the Stream Deck plugin; restart the Stream Deck app.
2. In an AI session as host: aim **Revoke Admin** at an AI car (viewed-car, specific number, or a Selected Car from the #732 selector) and press — the key flashes the warning triangle, nothing is sent, and the log shows `Refusing revoke-admin: target #N is an AI/pace car, not a user`. Your admin and session survive (the original report's repro).
3. Aim **Revoke Admin** at your own car — refused with `… is your own car`.
4. Enter a car number that isn't in the session on any of the five modes — refused with `… is not in the session`.
5. Aim **Black Flag / DQ / EOL** at an AI car — still sends normally.
6. Without iRacing: mock mode (`IRACEDECK_MOCK=1`) has AI cars (`CarIsAI: 1`) in its roster; chat commands are logged no-ops.

## Checklist

- [x] New code has unit tests

*(Mirabox/Ulanzi note: the guard lives in the shared action code and applies on all hosts; the `showAlert` key flash is Elgato-only — other adapters simply omit the optional method, and the refusal is still logged.)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Driver-management actions now refuse to send when the target is an AI car, pace car, or a car not present in the current session.
  * Revoke Admin also refuses revoking your own car.
  * Race-control commands continue to work for AI cars as before.
* **New Features**
  * Supported key actions can now display a warning indicator alert when an operation can’t be applied.
* **Documentation**
  * Updated changelog and Race Admin docs to explain the new target validation and warning behavior.


niklam:
Nitpick addressed in b9b5910b — `getPlayerCarNumberFromSessionInfo` had been inserted between `classifyCarNumberTarget`'s docblock and its declaration; each function now sits directly under its own doc. Comment-only move, `@iracedeck/iracing-sdk` rebuilt clean and its 32 session-utils tests pass.

Issue #753, opened 2026-07-04
Device-suffixed profile file names with clean user-facing names — codify the device-name mapping


niklam:
### What would you like?

Follow-up to #736 / #737 / #732 / #746. Adopt a device-aware naming scheme for bundled `.streamDeckProfile` files, while keeping the user-facing profile names clean:

- **File / manifest name** (`Profiles[].Name`, and the string passed to `switchToProfile`): `iRaceDeck <template> <device>` — e.g. `iRaceDeck Default XL.streamDeckProfile`, `iRaceDeck Race Admin Per Car Plus XL.streamDeckProfile`.
- **User-facing name** (the internal `Name` inside the exported bundle — what the Stream Deck app shows): stays `iRaceDeck <template>`, no device suffix. A plain file rename does not touch the ZIP contents, so the four existing XL bundles can be renamed without re-exporting.

**Device-name mapping** (canonical, must be codified in `packages/deck-core/src/device-profiles.ts` and written into the guides):

| Device | Suffix |
|---|---|
| Stream Deck (0) | `SD` |
| Stream Deck Mini (1) | `Mini` |
| Stream Deck XL (2) | `XL` |
| Stream Deck + (7) | `Plus` |
| Stream Deck Neo (9) | `Neo` |
| Corsair Galleon 100 SD (12) | `Corsair Galleon` |
| Stream Deck + XL (13) | `Plus XL` |

Rule: device names starting with "Stream Deck" use the latter part (`Plus` spelled out for `+`); the classic Stream Deck, whose latter part is empty, uses `SD`. The Galleon is the only planned non-"Stream Deck" device.

Because code switches profiles by the manifest name, the rename ripples:

1. **Rename the 4 bundled files** and update `manifest.json` `Profiles[]`; regenerate `data/profiles.json` (`pnpm generate:action-profiles`).
2. **`device-profiles.ts`** — add the suffix mapping (e.g. `profileDeviceSuffix(type)` / `profileFileName(template, type)`); `PROFILE_NAMES` stays the display-name source; `DEFAULT_SELECTOR_TARGET_PROFILE` and every `switchToProfile` call site resolve the device-suffixed name from the acting device's `deviceType` at switch time. Update tests.
[...]

PR #761, opened 2026-07-04
improve(actions): default Switch Profile to the default profile and lock borders off on profile/selector keys (#755)

### LINKED ISSUE #755: Profile-key polish: Switch Profile defaults to the default profile, no logo-key title, borders locked off on profile/selector keys
### What would you like?

Follow-up to #736 / #737 / #732 / #746. Three default-behavior polish items for the profile-related keys:

**1. Switch Profile defaults to the default profile, not "Select profile…".** A freshly-placed Switch Profile key currently has `profile: ""` — the PI shows the `ird-profile-select` placeholder, the key titles itself "SWITCH PROFILE", and pressing it is a no-op. Instead, default the selection to the device's **iRaceDeck Default** profile (device-resolved once the profile-naming issue lands): the PI dropdown pre-selects it and a press switches to it without any configuration.

**2. No default title on the default-profile key.** When the selected profile is the default one (or nothing is selected), the key shows the iRaceDeck logo — that's clear enough on its own, so the default title text must be empty (`profileTitle` currently renders "DEFAULT" / "SWITCH PROFILE"). Other profiles keep their existing labels (REPLAY, RACE ADMIN CARS, …). Users can still add a title via the normal Title Overrides.

**3. Border defaults off (and no glow) for Switch Profile and the select-car number keys.** These keys must render border-less even when the user has enabled the plugin-wide border defaults — a border+glow around a full page of big car numbers (or the logo key) defeats their clean look. Per-action overrides must still win, so this is the `locked` semantics that title/color metadata already have, extended to borders:

- Extend `parseIconBorderDefaults` / `resolveBorderSettings` in `@iracedeck/icon-composer` so icon `<desc>` border metadata supports `enabled` / `glowEnabled` defaults plus a `locked` array (resolution: per-action override → global *(skipped when locked)* → icon default → `BORDER_DEFAULTS`), mirroring `resolveTitleSettings`.
- Declare `{"border":{"enabled":false,"glowEnabled":false,"locked":["enabled","glowEnabled"]}}` in the Switch Profile icons (`packages/icons/switch-profile/*.svg`) and the dynamic selector template (`packages/iracing-actions/icons/race-admin-car-selector.svg`).

### Use case

The bundled profiles put these keys in front of every user: the Switch Profile key on `iRaceDeck Default` and 29 car-number keys per selector page. Today a new Switch Profile key does nothing until configured, the logo key carries a redundant label, and users with global borders enabled get 30 glowing rectangles per selector page. Sensible defaults make the shipped profiles look right with zero configuration.

### Area

Action (existing)

### Alternatives considered

- Hardcoding border-off in the two actions' code instead of icon metadata — works, but per-action border overrides in the PI would silently stop working there; the metadata + `locked` route keeps the standard override chain intact and is reusable by future icons.
- Making Switch Profile press fall back to the default profile while keeping the empty setting — fixes the no-op but leaves the misleading "Select profile…" PI state; defaulting the setting is cleaner.

### Additional context

- `resolveBorderSettings` currently reads only `borderColor` from icon `<desc>` metadata; `enabled`/`borderWidth`/`glowEnabled`/`glowWidth` pass `undefined` as the icon default and have no lock mechanism (`packages/icon-composer/src/title-settings.ts`).
- Item 1 interacts with the profile-naming issue: the default must resolve per device (`iRaceDeck Default <suffix>`), so implement after (or together with) the naming change; `SwitchProfile.onKeyDown`'s empty-profile guard and `profileTitle("")` both change.
- The `ird-profile-select` component gains a way to express a real default selection instead of only a placeholder.

**Affected artifacts**

- `packages/icon-composer` — `title-settings.ts` (`resolveBorderSettings`), `icon-template.ts` (`parseIconBorderDefaults`), tests
- `packages/icons/switch-profile/*.svg` + `packages/iracing-actions/icons/race-admin-car-selector.svg` — border metadata; regenerate previews (`generate-icon-previews.mjs`) and PI defaults (`generate-icon-defaults.mjs`)
- `packages/iracing-actions/src/actions/switch-profile/` — schema default, `onKeyDown`, `profileTitle`, `switch-profile.ejs` (dropdown default), tests
- `packages/pi-components` — `ird-profile-select` default-selection support
- `.claude/rules/icons.md` — document border `locked` metadata alongside the existing color/title locks
- Website — changelog entry; `docs/actions/communication/race-admin.md` if the selector defaults are described there

niklam:
## Related Issue

Fixes #755

## What changed?

Default-behavior polish for the profile-related keys, in three parts:

1. **Border locks in icon metadata** (`@iracedeck/icon-composer`): `parseIconBorderDefaults` now reads `enabled` / `glowEnabled` defaults and a `locked` array from `<desc>` border metadata, and `resolveBorderSettings` skips the global border defaults for locked fields — same semantics as the existing title locks; the per-action Border Overrides still win. All six Switch Profile icons and the Race Admin car-selector template declare `{"border":{"enabled":false,"glowEnabled":false,"locked":["enabled","glowEnabled"]}}`, so those keys stay border-less even when plugin-wide borders are enabled. Documented in `.claude/rules/icons.md`; previews regenerated.
2. **Switch Profile defaults to the default profile**: the `profile` setting defaults to `iRaceDeck Default` (schema) and an empty persisted value falls back to it at press time — a freshly-placed Switch Profile key switches somewhere useful with zero configuration and is never a no-op. The default-profile key (and an empty selection) renders the bare iRaceDeck logo with **no title**; other profiles keep their labels.
3. **No placeholder in the dropdown**: `ird-profile-select` gains a `default` attribute — the named profile is displayed as selected while the setting is empty (or holds a name not available on this device), display-only and never persisted. With a default there is no "nothing selected" state, so the "Select a profile…" placeholder option is not rendered at all — every choice in the list does something. Race Admin's Target Profile dropdown (no `default` attribute) keeps its meaningful placeholder ("use the bundled per-car profile").

## How to test

On a Stream Deck with the Elgato plugin built + installed:

1. Drop a fresh **Switch Profile** key: it shows the bare iRaceDeck logo (no title), the PI dropdown shows **iRaceDeck Default** selected (no "Select a profile…" entry), and pressing it switches to iRaceDeck Default.
2. Enable borders in **Border Defaults** (global): normal keys get the border; Switch Profile keys and the car-selector number keys stay border-less. A per-key **Border Overrides → Enable Border** still draws one on them.
3. Pick another profile in the dropdown: the icon/title update as before; picking **Back to previous** still works.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — the icon-composer border-lock resolution is shared by all plugins (glow stays force-disabled on Mirabox/Ulanzi via `__FEATURE_BORDER_GLOW__`); the Switch Profile action itself remains Elgato-only by design.

## Summary by CodeRabbit

* **New Features**
  * Added an iRaceDeck Default option to the Switch Profile action and profile selector, so empty/cleared selections now use the default profile automatically.
* **Bug Fixes**
  * Locked icon border fields now stay borderless even when plugin-wide borders are enabled, while per-key border overrides still apply.
  * Improved Switch Profile and profile selector behavior when saved profiles are missing or empty.
* **Documentation**
  * Documented the “locked” border-field behavior (including which fields can be locked) and its interaction with global border defaults.


niklam:
@coderabbitai Both nitpicks addressed in 1873d248:

1. **Hardcoded `default="iRaceDeck Default"`** — partially accepted. Wiring the compiled `PROFILE_NAMES` constant into the template isn't feasible: PI templates can only `require()` from the shared data dir (`packages/iracing-actions/src/actions/data/`), not from built packages, and a divergence would be display-only anyway — the action's press-time fallback to `PROFILE_NAMES.default` drives the actual behavior. Added an EJS comment tying the literal to the constant so a rename can't silently skip it; the #753 profile-naming rework will centralize these names properly.
2. **Locked `color`/width fields untested** — accepted. Added a locked-`color` test (icon default beats global) and a locked-`borderWidth`/`glowWidth` test (global skipped, `BORDER_DEFAULTS` fallback).

Issue #754, opened 2026-07-04
Race Admin car selector: populate cars with any number of keys per page, and eliminate the page-flip default-icon flash


niklam:
### What would you like?

Follow-up to #732 / #746. Redesign the Race Admin car selector's slot assignment so car population works with **any number of select-car keys per page — including a different count on every page** — and eliminate the default-icon flash when navigating selector pages.

**Current behavior (confirmed in `race-admin-selector.ts` / `race-admin.ts`):** a key's slot is computed purely from *(grid coordinates, full device grid, per-key `selectorPage`)*. `computeCarSlotIndex` assumes every non-reserved cell of the entire grid is a select-car key, reserves exactly the three corners, and advances pages by `carsPerPage = cols×rows − 3`. With any other layout this breaks:

- Fewer keys, or keys placed elsewhere: each key shows the car for its *cell position*, so empty cells become skipped, unreachable cars.
- A select-car key placed on a reserved corner is dead (blank icon, press is a no-op).
- Pages with differing key counts don't compress: page 1 on an XL always starts at slot 29 regardless of what page 0 actually holds.
- Devices with `grid: null` in `DEVICE_SPECS` — **including the Galleon 100 SD we plan to support** — get `carsPerPage = 0`, so the selector is entirely non-functional there.

**Runtime constraint that shapes the design:** the Elgato SDK only materializes action contexts for the *visible* page — `willAppear` fires per page visit, the payload has coordinates but no page index, and there is no API to enumerate actions on other pages (nor to `setSettings`/`setImage` on them, so "pre-setting" hidden pages is not possible). The bundled `iRaceDeck Race Admin Cars` profile (inspected) has 6 pages navigated by the built-in `com.elgato.streamdeck.page.previous` / `page.next` actions, and its keys carry per-key `Settings` baked at export.

**Layout-discovery options to evaluate** (they compose; decision in this issue):

1. **Parse our own bundled profile** — the plugin ships the `.streamDeckProfile` ZIP, so at startup it can read every page's select-car key positions and compute cumulative start slots per page. Handles non-uniform per-page counts exactly, keeps the Elgato-native page nav and the per-key `selectorPage` setting (to identify which page a key is on). Covers only iRaceDeck's bundled profiles.
[...]

PR #764, opened 2026-07-04
fix(actions): walk Back-to-previous through visited profiles with a default-profile fallback (#762)

### LINKED ISSUE #762: Switch Profile "Back to previous" never learns the starting profile — first Back does nothing (iRaceDeck Default missing from history)
## Bug

The Switch Profile "Back to previous" mode (#756) tracks the plugin's own named switches in `profile-switcher.ts`, but the profile the user is standing on **before the first plugin switch** — typically `iRaceDeck Default` — is never observed, so it never enters the history. Consequences:

- First press of Back finds no history and falls back to the app-level "no profile" pop, which only works immediately after a plugin-pushed switch and is consumed on use — anywhere else the app logs `Profile not found` and the press does nothing.
- Multi-hop flows (Default → Race Admin Cars → Per Car) toggle between the last two profiles and can never walk back to Default, because the history is only two deep (`current`/`previous`).
- Switches made from the PI "Stream Deck Profiles" accordion bypass the history entirely (`ElgatoPlatformAdapter.handleSendToPlugin` calls `switchToProfile` directly).
- The history is in-memory, so an app/plugin restart wipes it while the user stands inside a bundled profile.

The Elgato SDK offers no way to query or observe the active profile (verified against `@elgato/streamdeck` 2.1.0 — `profiles.switchToProfile` is the entire profile API), so the plugin must learn the active profile from its own artifacts.

## Fix

1. **Host-profile marker (detection).** New optional Switch Profile setting `hostProfile` ("This key is placed in profile", `ird-profile-select` dropdown fed by the existing `_deviceProfiles` push, empty default). On `willAppear`/`onDidReceiveSettings`, a key with a non-empty marker reports `notifyProfileVisible(deviceId, hostProfile)` to the profile switcher, so the plugin learns the active profile regardless of how the user got there (manual navigation, restart, app pop). Consider also wiring the marker into the Race Admin `select-car` keys so the selector pages report too.
2. **Stack-based history.** Replace the per-device `{current, previous}` pair with a stack (capped, e.g. 10): a named switch to a profile already in the stack unwinds to it, otherwise pushes; `notifyProfileVisible` no-ops when it matches the top, unwinds when present, pushes otherwise; Back pops and switches by name to the new top. Press-again-toggle behavior is intentionally replaced by walk-back.
3. **No-history fallback to Default.** With nothing to pop, Back switches by name to the device's bundled Default profile (resolved per device type via `profiles.json`; the action passes it, e.g. `requestProfileSwitchBack(deviceId, fallbackProfile)`), replacing the unreliable app-level pop.
4. **Record PI-accordion switches.** Route the `sendToPlugin` `switchToProfile` message through `requestProfileSwitch` so accordion switches enter the history.

## Manual authoring (maintainer, Stream Deck app)

- Set the host-profile marker on the Switch Profile keys in every bundled profile and adopt the Back-to-previous mode where intended (the shipped Replay back key is still a named switch authored with 1.23.0), then re-export all four bundles.

## Affected artifacts

- `packages/deck-core` — `profile-switcher.ts` (+ tests), export of `notifyProfileVisible`
- `packages/deck-adapter-elgato` — `handleSendToPlugin` routing (+ tests)
- `packages/iracing-actions` — `switch-profile.ts`/`.ejs` (+ tests); possibly `race-admin` select-car marker
- `packages/pi-components` — only if `ird-profile-select` needs a variant for the marker dropdown
- Bundled `.streamDeckProfile` files — re-exports (manual)
- `.claude/rules/profiles-and-devices.md` — Back-to-previous / history semantics
- Website — Switch Profile action page; `changelog.mdx` (fold into the existing unreleased Back-to-previous Features line, since #756 hasn't shipped)
- `docs/` action documentation + `iracedeck-actions` skill — new setting / mode behavior

## Acceptance

- Fresh start in Default → spoke to Replay → Back returns to Default (by name, no app pop).
- Default → Cars → Per Car → Back → Back lands on Cars, then Default.
- Back pressed with no history and no plugin-pushed switch pending switches to the device's Default profile instead of doing nothing.
- Accordion-initiated switches appear in the history.

niklam:
## Related Issue

Fixes #762

## What changed?

The Switch Profile "Back to previous" history never learned the profile the user started on (typically **iRaceDeck Default**) — the Elgato SDK offers no way to query the active profile — so the first Back press did nothing and multi-hop flows toggled between the last two profiles instead of walking back.

- **Stack-based history** (`deck-core/profile-switcher.ts`): the per-device `{current, previous}` pair became a stack (capped at 10). Named switches push; switching to a profile already on the stack unwinds to it; Back pops and switches by name, so Default → Cars → Per Car → Back → Back walks all the way home. Back passes no page, so the app restores the page you left (forward switches keep the #754 `page 0` behavior).
- **Host-profile marker**: new optional Switch Profile setting `hostProfile` ("Placed in profile" dropdown in the PI). A marked key reports its host profile via the new `notifyProfileVisible()` on appear and on settings changes, so the plugin learns the active profile even after manual navigation or a plugin restart.
- **Default fallback**: with nothing left to pop, Back switches to the device's bundled Default profile (resolved from `profiles.json`) and makes it the new history root; the unreliable app-level "no profile" pop remains only as a last resort when no Default exists for the device.
- **PI accordion switches recorded**: `ElgatoPlatformAdapter.handleSendToPlugin` now routes through `requestProfileSwitch` instead of calling the SDK directly, so "Stream Deck Profiles" accordion switches enter the history.
- **Re-exported bundles** (by @niklam): the Replay and Race Admin Cars back keys now use the Back-to-previous mode. Note: the bundles do **not** carry `hostProfile` markers yet — they were authored against a PI that predates the new dropdown. All current flows still behave correctly via the Default fallback (Default is the hub); a follow-up re-export with markers set will add restart/manual-navigation robustness.
- Docs: changelog feature line updated (walk-back + Default fallback), `profiles-and-devices.md` documents the stack semantics and adds the marker step to the authoring workflow.

## How to test

1. Fresh plugin start standing in iRaceDeck Default → press the Replay spoke → press Back in Replay → returns to Default (by name, first press).
2. Default → Race Admin Cars → select a car → Per Car → the named back key returns to Cars → Back in Cars → returns to Default.
3. Press Back with no prior plugin switch at all (e.g. right after a restart) → switches to iRaceDeck Default instead of doing nothing.
4. Switch profiles from the PI "Stream Deck Profiles" accordion, then press Back → returns to the profile you switched away from.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — Elgato-only feature; profile switching is a no-op on Mirabox/Ulanzi by design

## Summary by CodeRabbit

* **New Features**
  * Added a “Placed in profile” setting for Switch Profile actions to improve back-navigation behavior in bundled profiles.
  * “Back to previous” now steps through earlier visited profiles and can fall back to a default profile when history runs out.

* **Bug Fixes**
  * Improved profile tracking so navigation history is remembered per device.
  * Fixed back-navigation to handle more than just the last two profiles and better recover when users revisit profiles manually.


niklam:
@coderabbitai review

niklam:
@coderabbitai full review

Issue #771, opened 2026-07-05
fix(race-engineer): checkered flag callout fires when the flag comes out, not at the player's S/F crossing


niklam:
# Bug

The Race Engineer's checkered-flag callout ("Checkered flag" / "Qualifying is over") plays the moment iRacing raises the checkered flag — which is well before the player actually crosses the start/finish line. The callout must wait until the player takes the flag at the S/F line.

## User report (Discord, v1.23.0)

> I am not sure if this is a bug or not but when I do a race, it seems like the race engineer calls out the "qualifying is over" or "checked flag" well before I cross the finish line. For example, I just had "checked flag" at spa while I was at the beginning of the final chicane. — Tom B

> Yes, on ovals we get the checkered flag call out well before the S/F on the last Q lap. — Rafter

## Root cause

`packages/sim-events-iracing/src/diff/flags.ts` emits `flag.checkered.raised` on the rising edge of the `Checkered` bit in `SessionFlags`, and the `CHECKERED` scenario in `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.ts` speaks immediately. iRacing raises that bit the moment the session ends (qualifying clock expires, race leader finishes) — for everyone at once, which can be most of a lap before the player reaches the line.

## Expected behavior

The checkered callout fires when the player crosses the S/F line and actually takes the flag, in every session type that has per-driver flag-taking (qualifying, race — and practice when the driver is on a lap).

## Design notes

- **Prior art for S/F-crossing detection:** #657's `diff/pace-laps.ts` (forward `LapDistPct` accrual with a half-lap wrap threshold) and #569's `race.finished` latch (first `lap.completed` after the checkered raises in a race).
- **Prefer deferring the emission translator-side** (hold the raise in `diffFlags`/a sibling diff until the player's next S/F crossing) so `flag.checkered.raised` comes to mean "the player took the checkered" and the audio-scenarios wiring stays untouched. If a new event is introduced instead, the scenario-harness event template + shortcut must be added.
- **Don't hold forever.** If the player is not live on track when the checkered raises (in the garage, on pit road, towed, done for the session), speak promptly instead of waiting for a crossing that never comes. Reuse `isLiveOnTrack` / pit-road signals to decide.
- **Reset on session change** so a pending deferred checkered can't leak into the next session.
[...]

Issue #572, opened 2026-05-17
feat: Race Engineer position-change callout for invalid qualifying laps


niklam:
## Summary

Extend the qualifying position-change callout from #566 so it still fires after an **invalidated lap** — and prefixes the readout with *"That lap didn't count."* so the driver knows the time was thrown out.

**Current behavior (#566).** On `lap.completed` in qualifying, the Race Engineer announces:

- Improvement to P1 → *"That puts us on pole."*
- Improvement (not P1) or first valid lap → *"That puts us to P[n]."*
- Worsened, or unchanged on a non-PB lap → *"We're currently P[n]."*
- Unchanged on a PB lap → silent (lap-time-best already narrates).

The scenario doesn't currently distinguish valid vs invalid laps. An invalid lap that doesn't move the standings still produces *"We're currently P5."*, which is misleading — the driver hears a normal status readout for a lap that was actually thrown out.

**Proposed behavior.** When the just-completed lap is invalid:

- Always speak as "worse" framing (an invalid lap can't improve standings, and even if standings shifted from others' laps the user's own lap was the trigger). Skip the pole / better branches.
- Prefix with a new clip: *"That lap didn't count."* + the existing `position-intro-worse` clip + the existing `position-number` clip.

Result: *"That lap didn't count. We're currently P5."*

Reuses every existing position clip (`position-number/1.mp3`–`64.mp3` and `position-intro-worse/currently-01.mp3`). Only **one** new voice line is needed.

## Why now

- Removes a clear UX rough edge in the callout shipped in #566 — the driver currently gets the same readout whether their qualifying lap counted or not.
- Reuses the existing intro + number clips, so the cost is one ElevenLabs line plus a small payload/scenario edit.
- Establishes the pattern for "lap validity" as a first-class signal on `lap.completed` — future callouts (lap-time on invalid laps, sector-pace, etc.) will want the same flag.

## Architecture

Follows `.claude/rules/race-engineer-callouts.md`. Single-clip addition; modifies one bus payload, one diff module, one scenario file. No new bus event, no new family, no new per-callout opt-in (rides on `calloutEnabledPositionChange`).

### 1. Bus event — `@iracedeck/event-bus`

Extend the existing `lap.completed` payload in `packages/event-bus/src/event-catalog.ts` with one optional field:

```ts
"lap.completed": SimEvent<"lap.completed", {
  // ... existing fields (lap, lapTime, isBest, isFirstValid, position, …) …
  /**
[...]

PR #778, opened 2026-07-05
improve(race-engineer): two-stage white flag callout — heads-up at the flag, last lap at the line (#772)

### LINKED ISSUE #772: improve(race-engineer): two-stage white flag callout — heads-up at the flag, "last lap" at the S/F line
# Enhancement

The white-flag callout should become two-stage: a heads-up when the flag comes out, and the definitive "last lap" call when the player crosses the S/F line.

## Current behavior

`packages/sim-events-iracing/src/diff/flags.ts` emits `flag.white.raised` on the rising edge of the `White` bit in `SessionFlags`, and the single `WHITE` scenario in `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.ts` speaks immediately (with practice/qualifying/race pool variants). Like the checkered (see #771), the bit raises well before the player reaches the line — especially noticeable on ovals — so "last lap" plays while the player still has most of a lap to run before the last lap even starts.

## Desired behavior

Two callouts (wording indicative, not final — refine at implementation time):

1. **When the white flag raises:** a heads-up along the lines of "We're about to start the last lap."
2. **When the player crosses the S/F line under the white flag:** "This is the last lap."

## Design notes

- **Shares the player-S/F-crossing mechanism** with the checkered-timing fix (#771) — implement after (or together with) that issue so both ride the same detection.
- **Event shape:** either a second bus event for the crossing stage or a deferred second emission; whichever is chosen, the scenario-harness event template (`event-names.ts`, compile-time completeness check) and a shortcut button must follow if a new event is added.
- **Session-type semantics:** decide whether the two-stage split applies to all session types or race-only. In practice/qualifying the white flag means "session ends after this lap" and the existing single callout may remain the right shape there; the two-stage wording above is race-oriented. Keep the existing per-session pools for whichever stages retain them.
- **Opt-in:** prefer riding the existing `calloutEnabledFlagWhite` for both stages (the #572 "modifier of the same callout" precedent) rather than a second setting key, unless there's a reason to silence the stages independently.
- **Edge cases:** white raised while the player is not live on track (no crossing stage); white followed quickly by checkered (family preemption should keep this clean — both are `family: "flag"`); session change while the crossing stage is pending.
- **New voice lines:** new entries in every `packages/audio-assets/configs/<voice>.voice.json` (key parity enforced), omit `seed` or set it to `1` on new entries, scoped `generate --group`, manifest rebuild, clips committed.

## Affected artifacts

- [ ] `packages/event-bus` — only if a new event is added (catalog + value exports)
- [ ] `packages/sim-events-iracing` — flag diff + state (type **and** `createInitialState`) + tests
- [ ] `packages/audio-assets` — voice-config entries for all voices, generated clips, both manifests
- [ ] `packages/audio-scenarios` — pools, scenario(s), family wiring in `index.ts` if scenario ids change, tests (`flag-alerts.test.ts`, `register-pit-crew.test.ts`)
- [ ] `packages/deck-core` / PI (`pit-crew.ejs`) / both plugins' `plugin.ts` — only if a new opt-in key is introduced
- [ ] `packages/scenario-harness` — event template + shortcut if a new event is added
- [ ] Website changelog (`packages/website/src/content/docs/changelog.mdx`) — Improvements entry in the in-development section
- [ ] `.claude/rules/race-engineer-callout-examples.md` — reference entry for the two-stage pattern

niklam:
## Related Issue

Fixes #772

> **Stacked on #771's branch** (`ir-771`) so this PR shows only its own delta. When the #771 PR merges (head branch auto-deleted), GitHub retargets this PR to `master`; I'll sync the branch with master before merging.

## What changed?

The white-flag callout is now two-stage in **race sessions**:

1. **Heads-up at the raise** — iRacing shows the White to the whole field while the leader is still ~5–10 s from starting the final lap, so the raise precedes everyone's crossing. The `flag-white-race` clips were re-cut with heads-up wording (*"White flag. We're about to start the final lap."* + variant).
2. **Last-lap call at your crossing** — a new `flag.white-last-lap.raised` event fires at the player's own scored S/F crossing under the flag (*"This is the last lap. Give it everything."* + variant, new ElevenLabs clips), latched once per white episode.

Guards shaping stage 2:

- **Race-only in the diff AND the scenario** — in practice/qualifying iRacing shows the white exactly when the player starts their own final lap (same tick as their crossing), so a session-agnostic split would swallow the only callout those sessions get. They keep the single raise-time line.
- **`WHITE_LAST_LAP_MIN_GAP_MS` (6 s)** — a crossing landing while the heads-up is still playing (~4.5 s of audio) skips the second line instead of same-family-preempting it mid-sentence. Under real timing even the leader's crossing lands past the guard and gets both lines.
- **Checkered overlap guard** — the two bits can overlap for a tick on the finish crossing; the checkered call owns that crossing.
- **Leader skip** — if the bit ever lands at/after the race leader's own crossing, only the last-lap line speaks (safety net mirroring #771's winner grace).
- The scenario is **`queueable`** (the YELLOW_CLEARED precedent): a fire displaced by an equal-weight spotter call replays at idle instead of being dropped — the translator latch never re-fires.

Both stages ride the existing `calloutEnabledFlagWhite` opt-in (the #572 modifier precedent) — no schema/PI/plugin changes. The white latch + raise timestamp survive the replay-tick state wipe. Scenario-harness gets the new event template + a "White — Last Lap Started" shortcut.

## How to test

1. Timed race, mid-pack: heads-up when the flag comes out, *"this is the last lap"* as you cross S/F. Leading: heads-up ~5–10 s before the line, last-lap call at the line.
2. Cross within ~6 s of the flag coming out: only the heads-up plays (no talk-over).
3. Qualifying/practice: single raise-time white line, unchanged.
4. Scenario harness: "White — Last Lap Started" button fires the stage-2 line (race session preset).
5. Unit tests: `flags.test.ts` (white two-stage describe), `flag-alerts.test.ts` (scenario structure, race-only gating, queueable set).

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — no new opt-in key, existing PI checkbox covers both stages

## Summary by CodeRabbit

* **New Features**
  * Added a two-stage white-flag race announcement: a heads-up when the white flag appears, then a separate “last lap” callout at start/finish.
  * Added new audio lines and scenario coverage for the “last lap started” message.
* **Bug Fixes**
  * Prevented overlapping white-flag announcements and kept the second callout from being dropped when delayed.
  * Preserved the behavior across replay-like transitions.
* **Documentation**
  * Updated release notes and voice coverage docs to describe the new race announcement flow.


niklam:
@coderabbitai review

Issue #837, opened 2026-07-14
Race Engineer: acknowledgment pools config-driven (clip rename migration)


niklam:
## What would you like?

Phase 4 of #664: make the **acknowledgment pools** config-driven. Rename the curated acknowledgment clips to the `<base>-NN.mp3` convention so they follow the same derive-from-manifest rule as every other pool.

This is a **file-rename migration only** — rename the mp3s and rekey `generate.manifest.json`; the audio is identical, so there is **no** paid ElevenLabs re-cut. The `previous_request_ids`/`next_request_ids` chains in the voice configs reference clips by exact name, so every renamed base must have its refs updated.

## Use case

Acknowledgments are the last hand-enumerated pool family after phases 1–3. Once renamed, adding/removing an acknowledgment variant for any voice is just adding/removing a clip file — no code change.

## Affected artifacts

- **`@iracedeck/audio-assets`** — `configs/*.voice.json` (rename acknowledgment entries; fix `previous_request_ids`/`next_request_ids` refs), `voice/**` + `generate.manifest.json` (file-rename migration, no re-cut), `manifest.json` (rebuild).
- **`@iracedeck/audio-scenarios`** — remove any remaining acknowledgment-specific pool wiring in `catalog/pit-crew/` so the acknowledgments resolve through the standard registry.
- **Tests** — acknowledgment-related assertions in `register-pit-crew.test.ts` / family tests; generator reference-validation must pass after the rekey.
- **Docs** — `packages/audio-assets/CLAUDE.md` (acknowledgment naming convention), `.claude/rules/race-engineer-callouts.md` if it mentions acknowledgment clip names.
- **Website/changelog** — internal architecture; no changelog entry unless observable behavior changes.

## Risks

- **Request-id chains** reference clips by exact name; a missed ref fails the generator's reference validation — update every chain that touches a renamed clip.
- Rename must be a pure move (same audio bytes) so no re-generation cost is incurred.

## Area

Other (internal architecture / Race Engineer audio-assets + audio-scenarios).

## Related

Split out of #664 (phase 4 of the suggested phasing). Independent of the value-indexed-pools phase (#836); builds on phase 1's derive-from-manifest model (delivered under #664 itself).

Issue #836, opened 2026-07-14
Race Engineer: value-indexed clips as pools (drop hardcoded value ranges)


niklam:
## What would you like?

Phase 3 of #664: make the **value-indexed clips** (numbers, names, lap-time digits, temperatures, speeds) pools too — usually of size 1. The `var` resolvers return a pool reference instead of a raw clip path, and the engine skip rule (from the phase-2 issue) handles a voice that lacks a given value.

This removes the hardcoded bounds/speakability constants from code (`LAP_TIME_MINUTE_MAX`, `SESSION_START_TEMP_MAX`, `SESSION_START_SPEED_VALUES`, position max, …) — the clips that exist for a voice define what values it can speak.

To avoid renaming hundreds of number/temp/name clips, treat a bare `<base>.mp3` as a size-1 pool (preferred), or do a pure file-rename migration with **no** paid re-cut.

## Use case

The valid value ranges are currently duplicated between the voice config (which clips were generated) and code constants (which values are gated speakable). Deriving them from the manifest means extending a voice's range (or shipping a voice with a narrower range) is just adding/removing clip files — no code change, and no half-spoken callouts thanks to the engine skip rule.

## Affected artifacts

- **`@iracedeck/audio-scenarios`** — `dsl.ts` (`var` returns pool refs), `interpreter.ts` (bare `<base>.mp3` as size-1 pool if that route is chosen), and the value-composed families `{lap-time, position, position-readout, race-start, race-end, race-status, session-start, overtake, toggle-confirmations}.ts` (resolvers return pool refs; delete hardcoded ranges).
- **`@iracedeck/audio-assets`** — only if the rename route is chosen: `configs/*.voice.json` value entries, `voice/**` + `generate.manifest.json` (pure file-rename migration, no re-cut), `manifest.json` (rebuild).
- **Tests** — value-callout family tests: swap range-constant assertions to manifest-derived cases; add missing-value → skip-callout cases.
- **Docs** — `packages/audio-scenarios/CLAUDE.md`, `packages/audio-assets/CLAUDE.md` (value-clip convention), `.claude/rules/race-engineer-callouts.md`.
- **Website/changelog** — internal architecture; no changelog entry unless observable behavior changes.

## Risks

- **Rename scope** is large if the file-rename route is chosen — prefer bare-`<base>.mp3`-as-size-1-pool to avoid touching hundreds of clips and their `generate.manifest.json` keys.
[...]

Issue #835, opened 2026-07-14
Race Engineer: missing clip skips the whole callout (engine abort + optional clauses)


niklam:
## What would you like?

Phase 2 of #664: generalize the per-family speakability gates into **one engine rule** — during sequence expansion, every clip-producing step is checked against the runtime manifest for the active voice, and a required step that resolves to nothing **aborts the entire callout** (never a half-sentence).

Two genuinely-optional clauses (the pit-speed clause and the setup-warning clause in session/race-start) get an `optional` semantic so they skip locally instead of killing the callout.

**Ordering constraint:** the abort decision must happen **before** the bus scheduling/preemption step — a callout that would preempt an in-flight one but then aborts must not silence the in-flight callout.

## Use case

With pools derived per-voice from the manifest (#664 phase 1), a voice may omit clips or whole callouts. Without a general skip rule, a missing clip mid-sequence would either error or play a fragment. One engine rule replaces the scattered per-family `where:`/`if` speakability checks and guarantees a voice missing any required clip simply doesn't play that callout.

## Affected artifacts

- **`@iracedeck/audio-scenarios`** — `interpreter.ts` (expand-before-schedule + abort-on-empty), `dsl.ts` (`optional` clause semantic), family files marking their optional clauses (session-start, race-start).
- **Tests** — `interpreter.test.ts` and affected family tests: abort-before-preempt case, missing-clip → skip-whole-callout case, optional-clause-skips-locally case.
- **Docs** — `packages/audio-scenarios/CLAUDE.md` (engine rule + `optional` semantic), `.claude/rules/race-engineer-callouts.md` (missing → skip-callout rule).
- **Website/changelog** — internal architecture; no user-facing change expected, so no changelog entry unless observable behavior changes.

## Risks

- **Abort-before-preemption** is the subtlest change — the abort must be decided before scheduling so an aborting callout can't preempt (and thereby silence) an in-flight one.

## Area

Other (internal architecture / Race Engineer audio-scenarios).

## Related

Split out of #664 (phase 2 of the suggested phasing). Builds on phase 1 (pools derived per-voice from the manifest, delivered under #664 itself). Blocks the value-indexed-pools phase (#836).

PR #839, opened 2026-07-14
refactor(audio-scenarios): derive Race Engineer callout pools per-voice from the manifest (#664)

### LINKED ISSUE #664: Config-driven Race Engineer callout pools (derive per-voice from clips, drop hardcoded enumeration)
## What would you like?

Make Race Engineer callout **pools config-driven**, derived per-voice from the clips that actually exist, instead of hand-enumerated in code.

Today every pool is hand-listed in `packages/audio-scenarios/src/catalog/pit-crew/pools.ts` — each pool name maps to an explicit array of clip paths (e.g. `flag-blue → [voice/{voice}/flags/blue-01.mp3, …blue-02.mp3]`). Adding a variant means editing both the voice config **and** this code list; the code "knows" how many variants exist; and `voice-parity.test.ts` forces every voice to ship an identical clip set. The value-composed callouts (numbers, names, lap-time digits, temperatures) additionally hardcode their valid ranges in code (`LAP_TIME_MINUTE_MAX = 10`, `SESSION_START_TEMP_MAX = 150`, `SESSION_START_SPEED_VALUES`, position max) and gate on them via `where:`/`if` speakability checks.

**Target model:** the files that exist for a voice are the single source of truth. A "pool" is just *all clips sharing a base name* (`<base>-NN.mp3`), derived per-voice from the runtime manifest at fire time. The code never enumerates clips or knows counts. Different voices may carry different variant counts, or omit a callout entirely; a voice missing any required clip for a callout simply doesn't play **that whole callout** (never a half-sentence). This unifies the random-variant pools (flags, etc.), the acknowledgment pools, and the value-indexed clips (numbers/names/digits — each a pool, usually of size 1) under one rule.

## Use case

The pool enumeration is duplicated maintenance and a footgun: a clip and its pool entry can drift, and the rigid cross-voice parity means a new voice (or a voice that records fewer/more variants of a line) can't ship until it matches the canonical set exactly. We want to add variants, omit lines, or bring up a partial voice by **just adding/removing clip files** — with no code change and no half-spoken callouts.

## Approach (runtime-only)

Keep authoring entries with explicit names in the voice config (no generator auto-numbering or schema change). The change is in how the **runtime** resolves pools, plus a small clip-rename migration so the acknowledgment and value-indexed groups follow the `<base>-NN` convention. Generated filenames for the existing variant pools are unchanged, so those clips need no regeneration.

**1. Pool = `<group>/<base>`, members derived per-voice from the manifest.** Replace the `POOLS` clip enumeration in `pools.ts` with a thin `POOL_REGISTRY` mapping each family-prefixed logical name (`flag-blue`) to its `(group, base)` (`flags`, `blue`) — the names don't equal `<group>/<base>` (`flag`≠`flags`; `start-light-ready`→`start-lights/start-ready`), so a small map is needed, but it carries no clip lists or counts. Scenario reference strings (`"pool:flag-blue"`) stay unchanged. The interpreter resolves a pool's members at fire time by scanning the manifest's `clips[]` for `voice/<voice>/<group>/<base>-NN.mp3` using the already-known active voice. Delete the parity coupling in `definePool` that rejects a pool unless every voice has every clip; demote empty-pool from error to debug; reset the no-repeat `lastIndex` when the active voice changes (counts differ across voices). Validation keeps a typo guard (a base matching no clip for the reference voice warns, doesn't disable).

**2. Any required clip missing → skip the whole callout.** Generalize the per-family speakability gates into one engine rule and delete the hardcoded ranges. During sequence expansion every clip-producing step is checked against the manifest for the active voice; a required step that resolves to nothing aborts the entire callout (no fragment). Two genuinely-optional clauses (the pit-speed clause and the setup-warning clause in session/race-start) get an `optional` semantic so they skip locally instead of killing the callout. The abort decision must happen **before** the bus scheduling/preemption step, so a callout that would preempt an in-flight one but then aborts can't silence it.

**3. Acknowledgment pools config-driven.** Rename the curated acknowledgment clips to the `<base>-NN` convention (file-rename migration — rename mp3s + rekey `generate.manifest.json`, **no** paid re-cut since the audio is identical) and update the `previous_request_ids`/`next_request_ids` chains that reference them by exact name.

**4. Value-indexed clips as pools.** Numbers/names/temps/digits become pools (usually size 1); the var resolvers return a pool reference instead of a raw clip path, and the §2 rule handles a voice that lacks a given value. Removes the hardcoded bounds/speakability constants. (Recommend treating a bare `<base>.mp3` as a size-1 pool to avoid renaming hundreds of number/temp/name clips, or use a pure file-rename migration with no re-cut.)

**5. Relax `voice-parity.test.ts`** so voices may differ in variant counts and omit callouts (optionally keep a soft typo guard).

## Affected artifacts

- **`@iracedeck/audio-scenarios`** — `interpreter.ts` (pool resolution per-voice, expand-before-schedule + abort-on-empty, `definePoolFromManifest`), `dsl.ts` (`var`-returns-pool, `optional` clause), `validation.ts` (warn-not-disable typo guard), `catalog/pit-crew/pools.ts` (→ `POOL_REGISTRY`), `index.ts` (registration loop + `*_POOL_NAMES`), and the value-composed families `{lap-time, position, position-readout, race-start, race-end, race-status, session-start, overtake, toggle-confirmations}.ts` (resolvers return pool refs; delete hardcoded ranges; mark optional clauses).
- **`@iracedeck/audio-assets`** — `configs/*.voice.json` (rename acknowledgment, optionally value entries; fix request-id refs), `voice/**` + `generate.manifest.json` (file-rename migration, no re-cut), `manifest.json` (rebuild), `src/generate/voice-parity.test.ts` (relax parity).
- **Tests** — `register-pit-crew.test.ts`, `flag-alerts.test.ts`, `pit-status.test.ts`, `start-lights.test.ts`, `toggle-confirmations.test.ts`, `interpreter.test.ts`, and value-callout family tests: swap `POOLS[name]` assertions to registry/manifest; add cases for per-voice resolution, voice-change `lastIndex` reset, abort-before-preempt, and missing→skip-callout.
- **Docs** — `packages/audio-scenarios/CLAUDE.md` ("Pools" + "Adding a new family"), `packages/audio-assets/CLAUDE.md` (conventions/`-NN`), `.claude/rules/race-engineer-callouts.md` ("Audio pools" row + checklist step 4) to describe the derive-from-manifest model and the missing→skip-callout rule.

## Suggested phasing (each independently shippable)

1. Variant pools config-driven (§1 + §5 parity relaxation + docs). Deletes the bulk of `pools.ts`; zero scenario-reference edits; no clip renames. Delivers the core ask.
2. Missing → skip whole callout (§2 engine abort with the expand-before-schedule fix + `optional` clauses).
3. Value-indexed clips as pools (§4 + remove hardcoded ranges). Largest migration (renames).
4. Acknowledgment pools config-driven (§3 — renames + request-id rework).

## Risks

- **Abort-before-preemption** is the subtlest change — a callout that preempts then aborts must not silence the in-flight one; the abort must be decided before scheduling.
- **No-repeat tracker across voices** — reset `lastIndex` on voice change since counts differ.
- **Value-clip rename scope** is large — prefer bare-as-size-1 or a pure file-rename migration to avoid paid ElevenLabs re-cuts.
- **Request-id chains** reference clips by exact name; every renamed base must have its refs updated or the generator's reference validation fails.
- **Typo guard** now only covers the reference voice (deliberate parity relaxation) — a base misspelled for a non-default voice won't fail CI.

## Area

Other (internal architecture / Race Engineer audio-scenarios + audio-assets).

niklam:
## Related Issue

Fixes #664

## What changed?

Phase 1 of #664: Race Engineer callout pools are now **config-driven**, derived per-voice from the clips that exist in the audio-asset manifest, instead of hand-enumerated in code.

- `pools.ts` no longer lists any clips: `POOL_REGISTRY` maps each logical pool name (`flag-blue`) to its manifest `(group, base)` (`flags`/`blue`) and carries no clip lists or counts. Scenario reference strings (`"pool:flag-blue"`) are unchanged.
- New engine API `definePoolFromManifest(name, group, base)`: a pool's members are every `voice/<voice>/<group>/<base>-NN.mp3` in the manifest, resolved for the **active voice at fire time**. Voices may carry different variant counts or omit a callout entirely — an empty pool skips its step (debug, not error), and the no-repeat tracker resets on voice change since counts differ across voices.
- The all-voices parity coupling is deleted: `definePool` and scenario validation now check `{voice}`-templated paths against the **reference voice** only (`default`, else the first sorted voice) and **warn instead of rejecting/disabling** — a probable-typo guard without blocking per-voice divergence.
- The two acknowledgment pools (`acknowledgment`, `pit-action-acknowledgment`) stay enumerated in `ENUMERATED_POOLS` until the `-NN` rename migration (#837).
- `voice-parity.test.ts` relaxed to a base-level soft typo guard: a non-default voice may have fewer/more variants and omit callouts; only `<group>/<base>` keys unknown to the default voice fail (they'd never play — almost certainly a misspelling).
- Docs updated to the derive-from-manifest model: `packages/audio-scenarios/CLAUDE.md`, `packages/audio-assets/CLAUDE.md`, `.claude/rules/race-engineer-callouts.md`.

Follow-up phases split out of #664: #835 (missing clip → skip whole callout), #836 (value-indexed clips as pools), #837 (acknowledgment rename migration).

No changelog entry: internal architecture, nothing user-visible changes with the single shipped voice.

## How to test

1. `pnpm install && pnpm build && pnpm test` — full suite passes (6515 tests; new coverage for per-voice resolution, exact `<base>-NN` matching, voice-change no-repeat reset, empty-pool skip, and the reference-voice typo guards).
2. Audition without iRacing: `pnpm --filter @iracedeck/scenario-harness dev`, open `127.0.0.1:5750`, fire flag / pit-status / incident / start-light shortcuts — callouts play exactly as before (same clips, same random-variant behavior).
3. In iRacing: trigger a flag callout and a pit-service toggle; behavior is unchanged for the `default` voice.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — no plugin changes needed; the pool registration path is internal to `@iracedeck/audio-scenarios`

https://claude.ai/code/session_01NxubR68fSJmc7LpKLVC8Wn

## Summary by CodeRabbit

* **New Features**
  * Audio callout and pit-crew pools are now manifest-derived and resolve members for the active voice at playback time.
  * Voice clip parity is relaxed using `<group>/<base>` matching (variant `-NN` stripping), enabling differing clip sets per voice.
* **Bug Fixes**
  * Missing clips for non-reference voices no longer disable scenarios; issues now surface as warnings.
* **Documentation**
  * Updated audio asset and scenario guidance to reflect the new manifest-based pooling and flexible voice coverage rules.
* **Tests**
  * Expanded coverage for manifest-derived pools, active-voice behavior, and relaxed parity “typo guard” checks.

### DIFF
--- a/.claude/rules/race-engineer-callouts.md
+++ b/.claude/rules/race-engineer-callouts.md
@@ -41,7 +41,7 @@ voice scenario as the outermost short-circuit.
 
 | Concern | File |
 |---|---|
-| **Voice lines (source of truth)** | `packages/audio-assets/configs/<voice-id>.voice.json` — canonical: `default.voice.json`; key parity across voices enforced by `voice-parity.test.ts` |
+| **Voice lines (source of truth)** | `packages/audio-assets/configs/<voice-id>.voice.json` — canonical: `default.voice.json`; voices may differ in variant counts and omit callouts (issue #664), `voice-parity.test.ts` only guards against `<group>/<base>` keys unknown to default |
 | **Generated clips** | `packages/audio-assets/voice/<voice>/<group>/<name>.mp3` (gitignored locally; committed once stable) |
 | **Generator cache** | `packages/audio-assets/generate.manifest.json` |
 | **Runtime manifest** | `packages/audio-assets/manifest.json` (rebuilt by `generate:manifest`) |
@@ -49,7 +49,7 @@ voice scenario as the outermost short-circuit.
 | **Bus public exports** | `packages/event-bus/src/index.ts` (export new enums as values, not just types) |
 | **iRacing translator** | `packages/sim-events-iracing/src/diff/<name>.ts` + wired into `translator.ts` |
 | **Translator state** | `packages/sim-events-iracing/src/state.ts` (TranslatorState type AND createInitialState — keep them in sync) |
-| **Audio pools** | `packages/audio-scenarios/src/catalog/pit-crew/pools.ts` |
+| **Audio pools** | `packages/audio-scenarios/src/catalog/pit-crew/pools.ts` — `POOL_REGISTRY` maps pool name → manifest `(group, base)`; members (`<base>-NN.mp3`) derive per-voice from the manifest at fire time (issue #664) |
 | **Audio scenarios** | `packages/audio-scenarios/src/catalog/pit-crew/<family>.ts` |
 | **Family wiring (id type, key map, scenario id map, registerPitCrew param)** | `packages/audio-scenarios/src/catalog/pit-crew/index.ts` |
 | **Per-callout opt-in (Zod field)** | `packages/deck-core/src/global-settings.ts` |
@@ -62,7 +62,7 @@ voice scenario as the outermost short-circuit.
 
 - **Per-callout opt-in setting key:** `callout<Polarity><Family><Subject>` (`.claude/rules/global-settings.md` is the canonical reference). Polarity is always `Enabled`; the schema field's *default* encodes the family's natural baseline (callouts default `true`). Examples: `calloutEnabledFlagYellowLocal`, `calloutEnabledTrackWetness`.
 - **Scenario id:** `pit-crew.<family>-<subject>` — `pit-crew.flag-yellow-local`, `pit-crew.pit-status-too-far-left`, `pit-crew.track-conditions-worsening-mostly-dry`.
-- **Pool name:** `<family>-<subject>` — matches the scenario subject. One pool per scenario; arrays of clip paths so future variants append cleanly.
+- **Pool name:** `<family>-<subject>` — matches the scenario subject. One pool per scenario. A pool's members are all clips sharing its `(group, base)` — `voice/<voice>/<group>/<base>-NN.mp3` — so future variants are clip-file additions with no code change (issue #664). The pool name need not equal `<group>/<base>` (`flag-blue` → `flags`/`blue`); `POOL_REGISTRY` carries the mapping.
 - **Voice clip path:** `voice/{voice}/<group>/<name>.mp3` — group keys the `groups` map in the voice config; name keys the entry inside the group.
 - **Family identifier (for preemption):** matches the directory naming: `flag`, `damage`, `pit-status`, `track-conditions`. All scenarios in a family share the same `family:` value, so a newer fire supersedes an in-flight one cleanly.
 
@@ -86,7 +86,7 @@ In `packages/event-bus/src/event-catalog.ts`:
 
 ### 3. Voice lines
 
-In `packages/audio-assets/configs/default.voice.json` (and every other voice file — `voice-parity.test.ts` enforces matching `<group>/<entry-name>` sets across voices, even though text can differ):
+In `packages/audio-assets/configs/default.voice.json` (other voices *may* add the same entries with their own wording, but don't have to — a voice without a clip skips that callout; `voice-parity.test.ts` only rejects `<group>/<base>` keys unknown to default, issue #664):
 - Add (or extend) a group with one entry per `(direction × subject)` combination.
 - Each entry: `name` (kebab-case, suffix `-01` so future variants append as `-02`), `text`, optionally `seed` (omit it on new entries — the generator defaults an omitted seed to `1` — or set `"seed": 1` explicitly; NEVER an arbitrary/random value, since the seed only selects which take ElevenLabs produces. Bump it deliberately for a different take when the generated clip doesn't sound right — the seed feeds the hash, so the change re-cuts only that clip), optional `previous_request_ids` to bias prosody continuity.
 - Use `<break time="0.3s" />` for natural pauses inside a single line.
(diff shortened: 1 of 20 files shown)


niklam:
@coderabbitai Regarding the nitpick on `definePoolFromManifest` (stable `(group, base)` identifier):

**Adopted (docs):** a03e16df adds the stable-identifier semantics to the `definePoolFromManifest` JSDoc — renaming a base (in `POOL_REGISTRY` or on disk) silently changes or empties the pool and must be treated as a breaking change. This matches the convention already stated in `packages/audio-assets/CLAUDE.md`.

**Rejected (registration-time breaking error):** escalating a members-less pool to a rejection would contradict the #664 design on two counts. First, an empty pool is observationally indistinguishable from a legitimate omission — the spec explicitly allows a voice (including the reference voice, e.g. during a callout retirement) to lack clips, and chose "warns, doesn't disable" for exactly that reason. Second, the test-isolation pattern depends on it: family tests and `register-pit-crew.test.ts` register the full pool catalog against minimal fixture manifests, so most pools are legitimately empty for every fixture voice and must register cleanly and skip at fire time. The reference-voice warning remains the rename/typo signal; the whole-callout skip rule (#835) is the runtime safety net.

PR #841, opened 2026-07-14
improve(audio-scenarios): value-indexed clips resolve as manifest-derived pools (#836)

### LINKED ISSUE #836: Race Engineer: value-indexed clips as pools (drop hardcoded value ranges)

niklam:
## Related Issue

Fixes #836

## What changed?

Phase 3 of the #664 series: **value-indexed clips (numbers, names, temperatures, speeds, lap-time digits) are pools too** — usually of size 1 — and every hardcoded value range is gone. The clips that exist for a voice define what it can speak.

- **`poolRef(group, base)`** (new in `dsl.ts`): a `var` resolver returns a dynamic `pool:<group>/<base>` reference instead of a raw clip path. The interpreter derives its members from the manifest for the active voice at fire time (lazily cached, same no-repeat guard and voice-change reset as registered pools).
- **Bare `<base>.mp3` counts as a pool member** alongside `-NN` variants — for registered and dynamic pools alike — so the hundreds of existing value clips needed **no rename migration**, and any value clip can gain `-NN` variants later with no code change.
- **Hardcoded ranges deleted**: `SESSION_START_SPEED_VALUES` (+ findings tables), the temperature clamps (session-start and race-start `clampTemp`, `TEMP_MIN/MAX`), `POSITION_MAX` / `positionIsSpeakable`, `position-range.ts` (`POSITION_NUMBER_MIN/MAX`, `positionNumberIsSpeakable`), and `LAP_TIME_MINUTE_MAX` / `lapTimeIsSpeakable`. `where:` predicates and resolvers keep only null/known checks; the #835 abort/optional rules own speakability.
- **New optional clauses** where value-driven skipping is the right degradation: the session/race-start temperature clauses (an out-of-range reading now skips its clause instead of speaking a wrong clamped number) and the race-start position clause.
- Fixed lines (race-end result branches, intros, still-leading) are poolRefs too, so they can gain variants for free.
- Scenario-harness shortcut descriptions and the CLAUDE.md / rules docs updated to the derive-from-clips wording.

**Behavioral deltas worth knowing** (all deliberate):

1. Temps outside the generated clip range are no longer clamped to 0/150 — the temp clause skips. Speaking a clamped number was wrong information; the remaining brief still plays.
2. Values beyond the old bounds now speak **if a clip exists** (e.g. generate `lap-time-minute/12.mp3` and a 12-minute lap is announced) — widening a range is purely a generate-clips job.
3. A position readout for a value with no clip (e.g. P65 today) claims the shared 20 s readout cooldown in `where:` and then aborts at expansion. Fields beyond 64 cars are practically nonexistent, and the alternative (re-running side-effecting `where:` gates after expansion) would break the deferred-replay model — noted rather than engineered around.

## How to test

1. `pnpm install && pnpm build && pnpm test` — full suite passes (6539 tests; new coverage: bare-file pools, dynamic poolRef resolution/abort/optional, un-clamped temps, beyond-old-bound values speaking when staged).
2. Harness: `pnpm --filter @iracedeck/scenario-harness dev` → fire session-start / race-start / position / lap-time shortcuts — identical audio for the default voice; the P65 shortcuts stay silent (now via expansion abort).
3. In iRacing: session brief and a position change speak exactly as before.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — no plugin changes needed; internal to `@iracedeck/audio-scenarios` (+ harness copy text)

https://claude.ai/code/session_01NxubR68fSJmc7LpKLVC8Wn

## Summary by CodeRabbit

* **New Features**
  * Audio callouts now use manifest-driven “audio pools” to select numbered/value variants dynamically per active voice, with randomized selection and no immediate repeats.
  * Added a new pool reference capability so scenario variables can resolve to the appropriate pool at expansion time.

* **Bug Fixes**
  * Missing audio clips now cause required callouts to abort cleanly, while optional sections are skipped without partial announcements.
  * Removed hardcoded numeric bounds and temperature clamping so any value with a matching clip can be announced.

* **Documentation**
  * Updated guidance for optional mid-sentence steps and clarified pool semantics (including bare/base clips and variant pools).


niklam:
@coderabbitai Nitpick addressed — the three interpreter tests now build their references via `poolRef()` instead of hardcoded `pool:…` strings, so the builder↔parser contract is exercised. Suite green.

Issue #867, opened 2026-07-16
Spotter proximity calls ("Car left" / "Car right") can be silently dropped behind other callouts — they must always be heard


niklam:
### What happened?

A car pulled alongside but the spotter's proximity call was never heard — another Race Engineer line (a "fastest lap" readout or similar) occupied the voice bus and the "Car left"/"Car right" call was lost. A missed proximity call is the single most dangerous silence the engineer can produce: the driver may turn in on a car they were never told about.

### Expected behavior

Spotter proximity transitions must always be heard, immediately. No other callout — fastest lap, flags, pit chatter, even the CRITICAL-band lines — should be able to keep "Car left"/"Car right"/"Three wide" off the bus.

### Analysis

The spotter scenario (`packages/audio-scenarios/src/catalog/pit-crew/spotter-engine.ts`, `SPOTTER_CALL_SCENARIO`) is scheduled at `weight: WEIGHT.SAFETY (70)`, `interrupt: true`, `queueable: false`. Two consequences:

- The interrupt cut requires **strictly greater** weight (`interpreter.ts` `attemptFire`: `weight > runningWeight && entry.raw.interrupt === true`). Any in-flight **equal-weight** line (all flag callouts, pit-approach/pit-exit/stall-departure — all `WEIGHT.SAFETY`) or **higher-weight** line (meatball, fuel-critical "Box this lap", start gantry — `WEIGHT.CRITICAL`) blocks the spotter call, and since it's `queueable: false` (correct — a stale proximity call must never replay late) it is **silently dropped**.
- The reported case ("fastest lap", i.e. the lap-time family at default `WEIGHT.NORMAL (50)`) should in theory already lose to SAFETY + interrupt — reproduce and investigate whether an engine path (pending-slot interaction, radio-frame include, still-there loop timing, focus-release timing) drops the call even against lower-weight lines. There may be a scheduler bug in addition to the weight-band hole.

### Proposed direction

- Add a band **above CRITICAL** to `WEIGHT` in `dsl.ts` (e.g. `PROXIMITY = 120`) and move the spotter call onto it with `interrupt: true` kept. Strictly-above is required — putting it AT `CRITICAL` still drops it against an in-flight CRITICAL line because equal weight never cuts.
- Keep `queueable: false` — the guarantee must come from winning the bus immediately, never from deferred replay.
[...]

Issue #671, opened 2026-06-11
Waving-yellow callouts need rewording, and a waving yellow must announce "Yellow cleared" when it ends


niklam:
### What happened?

The #480 flag rework (shipped after v1.20.0) split `YellowWaving` / `CautionWaving` out of the static yellow callouts into their own lines. Two problems with the result:

1. The new waving callout wording needs to be reworked. Current lines: `yellow-waving-01` = "Local yellow waving. <break> Slow, hazard ahead!", `caution-waving-01` = "Caution coming out!".
2. When a waving yellow ends, no "Yellow cleared." callout plays. Before the rework (v1.20.0), a waving yellow was treated as a plain yellow, so clearing it always announced "Yellow cleared."

### Expected behavior

- Waving-yellow lines reworded (final wording to be decided in this issue).
- A waving yellow (yellow-waving or caution-waving) that ends is followed by the "Yellow cleared." callout, same as a static yellow — restoring the v1.20.0 behavior.

### Steps to reproduce

1. Race with the Race Engineer enabled on current master (post-v1.20.0).
2. Trigger a waving local yellow (incident in your sector) — hear "Local yellow waving. Slow, hazard ahead!"
3. Let the yellow end — no "Yellow cleared." is heard.

### Additional context

- Detection: `packages/sim-events-iracing/src/diff/flags.ts` — `flag.yellow.cleared` is intended to fire when all yellow-ish bits clear (`state.lastAnyYellow`), so the missing clear needs investigation (translator edge vs scenario layer vs real-session bit behavior).
- Scenarios: `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.ts` (`YELLOW_WAVING`, `CAUTION_WAVING`, `YELLOW_CLEARED`).

Affected artifacts to keep in sync:

- Voice configs for **all voices** (`packages/audio-assets/configs/*.voice.json`, parity enforced) + regenerated clips + both manifests (`generate.manifest.json`, `manifest.json`).
- `packages/sim-events-iracing/src/diff/flags.ts` + tests, if the cleared edge is the bug.
- Scenario-harness shortcuts unaffected unless events change.
- No PI/schema changes expected (existing `calloutEnabledFlagYellowWaving` / `calloutEnabledFlagCautionWaving` / `calloutEnabledFlagYellowCleared` keys stay).

Issue #669, opened 2026-06-08
feat(audio-scenarios): Black Flag Furled — 1s debounce on raised + new "furled cleared" callout


niklam:
### What's wrong / what we want

The Race Engineer announces the **Black Flag Furled** warning immediately when iRacing raises the `Furled` session-flag bit. Running briefly off track flashes that bit for ~0.5 s even though it isn't a real furled-black-flag warning, so the engineer gives the furled readout when it shouldn't. Two changes, together:

1. **Debounce the raised callout by 1 second.** Only announce once the `Furled` bit has stayed high for ≥ 1 s; if it clears within that second, drop the announcement (nothing plays).
2. **Add a paired "furled cleared" callout** that fires when a *genuine* (announced) furled flag is withdrawn — mirroring the existing `flag.yellow.cleared`. It is gated on the raised callout having fired, so a transient flicker fires **neither** raised nor cleared.

### Steps to reproduce (the bug)

1. In a race, drive off track briefly so iRacing flashes the furled bit.
2. Today the engineer plays the furled callout despite no genuine warning.

### Root cause

`flag.furled.raised` is a plain rising edge in `packages/sim-events-iracing/src/diff/flags.ts` (`case "furled"` in the "new raised transitions" switch), with no time gate. `diffFlags(...)` doesn't receive `now` today (unlike `diffDamage` / `diffLimiter` / `diffToggles`). The only paired "cleared" event today is `flag.yellow.cleared` (falling-edge of `lastAnyYellow`).

### Fix

Reuse two existing patterns, scoped to the furled bit only:

- **Debounce** like `diff/damage.ts` (`DAMAGE_DEBOUNCE_MS` / `damagePendingAt`): thread the per-tick `now` into `diffFlags`; add `FURLED_DEBOUNCE_MS = 1000` (exported for tests); add `furledPendingAt` + `furledAnnounced` to `TranslatorState` (type **and** `createInitialState`). On a genuine `Furled` rising edge start the timer (don't emit); emit `flag.furled.raised` only once the bit has stayed set ≥ 1 s (set `furledAnnounced`); reset the timer the moment the bit clears.
[...]

PR #932, opened 2026-08-06
fix(audio): announce black flag when it lands during another callout (#923)

### LINKED ISSUE #923: Race Engineer: black flag can go unannounced when it lands during another callout — make the black-flag scenario queueable
## Symptom

A directly-issued black flag (no furled warning phase — e.g. a pit-lane speeding penalty, a race-admin `!black`, or escalation after ignoring the meatball) can go completely unannounced. If the Voice bus is busy with an equal-weight (`WEIGHT.SAFETY`) or higher-weight line at the instant `flag.black.raised` fires, the fire is dropped permanently. The raise is a one-shot edge that never re-fires, so the driver is never told they have a penalty.

The clearest real-world case: a spotter proximity call ("Car left.", `WEIGHT.PROXIMITY`) is playing when the black flag lands. The black flag MUST still be announced after the spotter clip ends — today it is silently lost.

## Root cause

`BLACK` in `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.ts` runs at `WEIGHT.SAFETY` with the default `queueable: false`. Interpreter scheduling (`packages/audio-scenarios/src/interpreter.ts`): an arriving fire at equal/lower weight than the in-flight one goes through `queueOrDrop` and, without `queueable`, is dropped with only a debug log.

Collisions with lower-weight lines already resolve correctly — a higher-weight non-interrupt fire waits as pending and plays when the current line finishes — so the speeding-warning and incident-points callouts (both `WEIGHT.NORMAL`, the lines most likely to be talking when a direct black flag lands) are not the problem. The loss is only against SAFETY-and-above occupants: other flag lines, pit-approach / pit-exit / stall-departure, the spotter still-there info line (SAFETY), meatball / fuel-critical / start gantry (CRITICAL), and spotter proximity calls (PROXIMITY).

The detection side is fine: `diffFlags` emits `flag.black.raised` on the plain rising edge with no debounce and no furled dependency (test-covered in `flags.test.ts`).

## Fix

Same class as #671 (yellow-cleared) and #867 (meatball, start-gantry lines): the penalty is a sustained state, so a replay a few seconds late is always still correct.

- Set `queueable: true` on `BLACK`.
- Consider a speak-time `if:` gate re-checking the live `Black` bit (the FURLED pattern from #669, including the missing-telemetry fail-open convention) so a queued fire that was overtaken — penalty served, or escalated to `Disqualify` — expands to silence instead of announcing a stale flag. Decide during implementation whether the queue latency (seconds) justifies it.
- Evaluate `DISQUALIFY` and `DQ_SCORING_INVALID` for the same treatment — identical one-shot penalty edges at SAFETY with the same loss paths, and a DQ is arguably even more must-hear.

## Repro

Scenario harness: fire any SAFETY/CRITICAL line (e.g. a spotter call or the meatball), then fire `flag.black.raised` while it is still playing. Expected: the black-flag line plays when the bus idles. Actual: dropped (debug log `Scenario "pit-crew.flag-black" dropped (bus busy)`).

## Affected artifacts

- `packages/audio-scenarios/src/catalog/pit-crew/flag-alerts.ts` — scenario metadata (+ `flag-alerts.test.ts`; `register-pit-crew.test.ts` if its fire matrix asserts scheduling flags)
- `packages/audio-scenarios/CLAUDE.md` — keep the queueable one-shot roster accurate (it currently names the #867 meatball/gantry set)
- `.claude/rules/race-engineer-callout-examples.md` — extend the reference catalog on merge
- Changelog `packages/website/src/content/docs/changelog.mdx` — user-facing bug fix, one line
- No plugin, manifest, PI, schema, event-bus, or scenario-harness changes (no new event, no new opt-in)

niklam:
## Related Issue

Fixes #923

## What changed?

A directly-issued black flag (pit-lane speeding penalty, race-admin `!black`, escalation after an ignored meatball) could go completely unannounced: the `BLACK` scenario ran at `WEIGHT.SAFETY` with the default `queueable: false`, so a raise landing while the Voice bus was busy with an equal- or higher-weight line was dropped permanently — the raise is a one-shot edge that never re-fires.

- `BLACK`, `DISQUALIFY`, and `DQ_SCORING_INVALID` are now `queueable: true` — a fire that can't take the bus defers and replays when the bus idles. Same class as #671 (yellow-cleared) and #867 (meatball/gantry): the penalty is a sustained state, so a seconds-late replay is always still correct.
- Deliberately **no** speak-time `if:` gate (the FURLED #669 pattern): serving a black flag takes a pit visit (≥ 30 s), far beyond queue latency, and the black→DQ escalation case is covered structurally — the now-queueable DISQUALIFY fire replaces the pending black in the single pending slot (equal weight, ties → newest). Design rationale in `docs/superpowers/specs/2026-08-05-black-flag-queueable-design.md`.
- Tests: roster assertion extended; per-scenario defer-and-replay tests; pending-slot escalation test; and the #867-pattern mid-playback supersession guard (a black line cut by the escalating DQ is never stashed for replay).
- Docs: `packages/audio-scenarios/CLAUDE.md` queueable roster, new `## 2.4.0` changelog section (Bug Fixes), and a #923 entry in `.claude/rules/race-engineer-callout-examples.md`.

No plugin, manifest, PI, schema, event-bus, clip, or scenario-harness changes.

## How to test

1. `pnpm --filter @iracedeck/scenario-harness dev`, open `http://127.0.0.1:5750`.
2. Fire a spotter proximity call (or the meatball), and while it is still playing fire **Black flag**. Expected: the black-flag line plays as soon as the current clip ends (previously: silently dropped with only a debug log).
3. Repeat with **Disqualify** and **DQ scoring invalid**.
4. Escalation: while a line plays, fire Black flag then Disqualify — only "Disqualified. Pull off." plays.

Verified manually in the harness; full suite green (7427 tests).

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — n/a, no plugin-side changes

https://claude.ai/code/session_014AovdhELJCq1Vk9s5trnQs

## Summary by CodeRabbit

* **Bug Fixes**
  * Race Engineer black-flag, disqualification, and scoring-invalid penalty alerts now wait when the voice channel is busy and play once it becomes available.
  * Newer disqualification alerts replace outdated queued black-flag notifications during escalation.

* **Documentation**
  * Added release notes and guidance describing the updated penalty-alert delivery behavior.


niklam:
Both nitpicks addressed in 1115fe9e:

1. **Mid-playback supersession test** — now advances past the radio-open tick before the escalation lands, so the black voice clip is genuinely in flight when the DQ cuts it, and the assertion tightened to played-exactly-once. One deviation from the proposed diff: the advance is `_triggerChannelEnd(AudioChannel.SFX)`, not `AudioChannel.Voice` — the sequence's first op is the walkie tick on the SFX channel, so no Voice callback is registered yet at that point and ending Voice would be a no-op; completing the SFX tick is what starts the black voice clip.

2. **Higher-weight blocker path** — `defineBlocker` now takes a weight and a parameterized case runs all three penalty events behind a `WEIGHT.CRITICAL` blocker, asserting the same idle replay.

Issue #880, opened 2026-07-19
Timed-race laps-remaining estimate for fuel-callout race-coverage suppression (follow-up to #866)


niklam:
### Background

Issue #866 added race-coverage suppression to the laps-of-fuel-left callout family, but only lap-limited races get the full estimate-vs-race-length comparison — timed races (where `SessionLapsRemainEx` reads the `IRSDK_UNLIMITED_LAPS` sentinel) only get the final-lap (white-flag crossing latch) and post-race gates. The #866 code review identified three further limitations that were explicitly accepted for the bug fix and deferred here.

### Scope

**Main feature — timed-race laps-remaining estimation.** Estimate the remaining race laps from `SessionTimeRemain` and an average lap time so the coverage comparison also works in timed races. Needs a lap-time tracker (a rolling clean-laps average analogous to the #465 fuel tracker — out-laps/in-laps/tow laps excluded) and a decision on leader pace: a timed race ends when the LEADER's clock expires plus their final crossing, so a mid-pack player's remaining laps depend on the leader's lap time, not just their own.

### Known limitations from the #866 review to resolve alongside

1. **Dual time+lap limits:** in sessions configured with BOTH a time and a lap limit, `SessionLapsRemainEx` reads a real lap count while the time limit may end the race sooner, so the #866 comparison overstates the remaining distance and the pre-#866 unnecessary callouts persist in that format. The time-based estimate should take the minimum of the two remaining distances.

2. **`SessionLapsRemainEx` race semantics unvalidated:** the #866 comparison assumes the counter is player-relative in races (proven only for qualifying, #776). If it is leader-relative, a lapped driver's remaining distance is understated and suppression can engage slightly early. Validate from a race telemetry capture while running laps down (the `local/telemetry-snapshot-*.json` precedent, #604/#647) and adjust the comparison if needed.

3. **White-flag latch resets on a final-lap flag change:** `whiteLastLapFired` re-arms to false the moment the White bit drops (flags.ts), e.g. when a caution replaces the white during the final lap of a timed race — removing the only final-lap guard in that format, so the box call can fire mid-final-lap under caution. Consider a sticky per-session "player started their final lap" latch (a `TranslatorState` field preserved across the replay wipe) if the timed estimate doesn't subsume the case.

### Affected artifacts

[...]

Issue #647, opened 2026-06-04
Parade-lap position changes are called out and displayed before the green flag


niklam:
## What happened?

Before the green flag — during the grid / formation / parade lap of a race — the field's running order churns as cars settle into formation. Two features react to that pre-green churn instead of holding the qualifying grid order:

1. **Position-change callouts fire on the parade lap.** The overtake diff (`packages/sim-events-iracing/src/diff/overtakes.ts`) gates on `isRaceSession`, `onPitRoad`, and `underCaution` — but **not** on `SessionState`. Parade-lap shuffling isn't flagged as a caution, so the Race Engineer announces phantom "Nice pass / we lost a position" callouts before the race has even started.
2. **The position display churns on the parade lap.** The Session Info action's position mode (`packages/iracing-actions/src/actions/session-info/session-info.ts`, ~L460-481) uses the calculated live running order (`getLivePosition()`) for any race session not on pit road, so the displayed P-number jumps around during the parade lap instead of showing the driver's qualifying grid slot.

These are the same root cause — consuming the live running order before the race is green — so they're tracked together.

## Expected behavior

- **No position-change callouts** are made until `telemetry.SessionState` reaches `StateRacing` (`SessionState.Racing` = `4`). The overtake baseline freezes/seeds silently pre-green (same as it already does under caution), so the first detection after green flies is measured from the settled green-flag order — no phantom callout at the green.
- **The position display shows the qualifying position** (the official grid slot — `PlayerCarPosition` / `PlayerCarClassPosition`) until that same moment (`SessionState === Racing`). Once racing, it resumes the live calculated order as today.

## Steps to reproduce

1. Qualify for a race so you start from a known grid slot.
2. Watch the grid / formation / parade lap before the green flag, with the Race Engineer enabled and a Session Info action in position mode.
3. Observe position-change callouts firing and the displayed P-number changing during the parade lap, before `SessionState` is `Racing`.

## Implementation notes

- `SessionState` is already on `TelemetryData` and `SessionState.Racing === 4` (`packages/iracing-native/src/defines.ts`). `diffOvertakes` already receives `telemetry`, so the gate can read `telemetry.SessionState` directly — AND it with the existing gates and roll the baseline silently while pre-green (mirroring the `underCaution` branch) so the green-flag order seeds the baseline.
[...]

PR #997, opened 2026-08-16
improve(race-positions): hold the grid order through the parade lap so positions exist before the green (#974)

### LINKED ISSUE #974: improve(race-positions): hold the grid order through the parade lap so positions exist before the green
### What would you like?

Race positions should **hold the starting-grid order throughout the parade lap**, so that:

- every position-aware display shows a sensible number before the green rather than a blank, and
- the Camera Controls dial's **Cycle by Race Position** mode can cycle the field during the parade lap (today it does nothing at all).

### Use case

Spectating / broadcasting the formation lap. Right now the Position dial is inert until the field completes a lap, so the one mode built for "show me P1, now P2, now P3" is unusable during the exact phase where a director is setting up shots.

### Current behavior (measured)

From capture `local/telemetry-snapshot-20260815-203305-955.json` — an AI race, `SessionState = 3` (`ParadeLaps`), 20 competitors plus the pace car at carIdx 64:

- `CarIdxPosition` is **0 for every car** — the official counters are empty pre-green.
- The canonical order is **all zeros** too: `calculateRacePositions` (`packages/iracing-sdk/src/position-utils.ts`) only ranks cars with `CarIdxLapCompleted >= 0`, and nobody has completed a lap yet, so `active` is empty.

So `resolveOrder()` yields zeros by either path → `computeRacePositionTarget` sees `maxPosition = 0` → returns `null` → the dial dispatches nothing.

### The trap: lap distance is NOT the grid order

The obvious fix — relax the lap-count gate and rank by `CarIdxLapDistPct` — produces a **subtly wrong** order, and the capture proves it. The parade lap is double-file, so within every row the outside car sits marginally further around the lap. Ordering by lap distance swaps **every single pair**:

| Grid (`QualifyResultsInfo`) | By `CarIdxLapDistPct` |
|---|---|
| P1 `#13`, P2 `#15` | `#15`, `#13` |
| P3 `#09`, P4 `#777` | `#777`, `#09` |
| P5 `#78`, P6 `#2` | `#2`, `#78` |
| P7 `#47`, P8 `#32` | `#32`, `#47` |
| … every remaining pair likewise swapped … | |

So the grid order must come from **`QualifyResultsInfo`**, not from telemetry geometry.

### Suggested approach

Add a pre-green grid source to the canonical order, gated to exactly the window where it applies:

- `isPreGreen(telemetry)` (`@iracedeck/iracing-sdk`) already covers `Invalid` / `GetInCar` / `Warmup` / `ParadeLaps` — the right gate, already used by the #647 overtake suppression.
- `extractQualifyResults(sessionInfo)` (`irating-utils.ts`) already parses `QualifyResultsInfo.Results`; its doc says it exists "so the shape knowledge lives in one place", so reuse it rather than re-parsing.
- `orderFromQualifyResults` (same module) already converts the 0-indexed `Position` field to 1-based ranks with validity rules for malformed YAML. Verified against this capture: positions are contiguous `0..19` for 20 cars, and every `FastestTime` is `-1` (an AI race with no qualifying — the grid is still populated).

**Note this crosses a boundary the rules currently draw.** `.claude/rules/race-positions.md` records the qualifying grid as an **estimate-only** input used by the iRating feature, with "the canonical position order and its consumers are unaffected". This issue deliberately changes that, so the rule must be updated in the same change.

### Risk assessment

- **Spurious overtake callouts: already handled.** `diffOvertakes` suppresses the pre-green states (#647) precisely because "grid / warmup / parade-lap shuffling" leaked position-change callouts before the green. A static grid order would produce no diffs anyway, but the existing gate makes this safe by construction.
- **The green-flag transition needs thought.** At green the order switches from held-grid to live lap-progress, and the two disagree pairwise (above). Whatever consumes the order across that boundary must not read the switch as a burst of position changes. Worth checking `diffGaps` and the Race Engineer position callouts alongside overtakes.
- **Blast radius.** The canonical order feeds Session Info Position and Gaps, Race Engineer position callouts, opponent-pit and opponent-flag classification, gap tracking, and the camera/keypad position modes. Each should be sanity-checked pre-green, not just the dial.

### Open questions

- **Cars not in the world.** In this capture the player (carIdx 0) holds grid `Position 0` but is not on track (`IsOnTrack = false`, no lap distance). Should a grid slot count when its car is absent? The camera walks skip absent cars via `carInWorld`, so the dial would step past it — but the displayed field size would still include it.
- **Class positions.** `QualifyResultsInfo` carries `ClassPosition` too; decide whether the pre-green order feeds `classPositionFromOrder` or whether class comes from the grid entry directly.
- **Which pre-green states.** `isPreGreen` includes `Warmup` and `GetInCar`. Holding the grid order that early is probably right, but confirm it doesn't surprise a practice/warmup display.
- **Rolling vs standing starts, and multi-class fields** — both worth a capture before shipping.

### Affected artifacts

- [ ] `@iracedeck/iracing-sdk` — the pre-green grid source (`position-utils.ts` and/or the translator's `calculateFrozenRacePositions`), reusing `extractQualifyResults` / `isPreGreen`; tests from the real capture above
- [ ] `@iracedeck/sim-events-iracing` — `race-finish.ts` position tracking and the green-flag transition; verify `diffOvertakes` / `diffGaps` behave across it
- [ ] **Rules** — `.claude/rules/race-positions.md` (the canonical-source section and the estimate-only boundary quoted above)
- [ ] **Website** — `changelog.mdx`; any action page that documents positions being unavailable before the green
- [ ] No new action, mode, binding, manifest, or comms-catalog entry expected

### Alternatives considered

- **Rank by lap distance pre-green.** Rejected — the capture shows it swaps every grid pair on a double-file parade lap.
- **Leave positions empty and rely on Cycle by Track Order** for parade-lap car hopping (which already works, #886). Viable, but it does not give the Position dial or any position display a value before the green.

### Additional context

Found while testing #968 (formation-lap car cycling). #968 fixes Cycle by Car # on the parade lap; it deliberately does **not** touch the position order, which is this issue.

niklam:
## Related Issue

Fixes #974

## What changed?

Before the green flag of a race the canonical race order ranked nobody. It scores cars by `CarIdxLapCompleted + CarIdxLapDistPct`, and iRacing holds that counter at `-1` until a car crosses start/finish (#307); its own `CarIdxPosition` counters are all-zero for the same reason. So every position readout was blank and the Camera Controls dial's **Cycle by Race Position** mode did nothing during exactly the phase a broadcaster is setting up shots.

The order now takes the **qualifying grid** as its pre-green source.

- **`packages/iracing-sdk/src/grid-utils.ts`** (new) owns the `QualifyResultsInfo` shape in one place: `extractQualifyResults` (moved out of `irating-utils.ts`) plus `calculateGridPositions`, the exported and hardened form of the private `orderFromQualifyResults`. It additionally drops a second car claiming an already-taken grid slot, so the grid source keeps the no-duplicate-ranks invariant `findDriverByRacePosition` relies on.
- **`packages/sim-events-iracing/src/race-order.ts`** (new) makes the source choice: the grid while `isPreGreen` holds in a race session, the frozen lap-progress order otherwise. The translator's single `resolveCanonicalOrder` derives `isRaceSession`, so `getLiveRacePositions`, `getLiveCarPosition`, `resolvePlayerIsLeader` and `handleTick` cannot diverge.

No consumer needed a code change — the camera dial, template variables, Session Info, the iRating estimate and the audio readouts all read the canonical getters and simply start receiving a real order pre-green.

### Two gates are load-bearing, both decided from real captures

**The gate is the session phase, not "the live order is empty."** By the end of a rolling parade lap the leaders have already crossed S/F for the pace lap and *do* carry a lap-progress score — capture `20260624-214144`: two of 21 cars, ranked in the **reverse** of their grid slots. An emptiness test would hand over to a two-car inverted order mid-formation and blank the other nineteen. There is a test pinning exactly that.

**It is race-only.** `SessionState` sits at `Racing` for the whole of practice and qualifying, so the pre-green states appear there only on transient session-boundary ticks — where `QualifyResultsInfo` holds the previous session's grid, or a qualifying order still being set.

Ranking the formation by lap distance is not an option either: a parade lap is double-file, so the outside car of every row sits marginally further round the lap and a distance ranking swaps **every** grid pair.

### Decisions worth flagging

- **Cars not in the world keep their grid slot.** A driver still in the garage starts from their qualifying position all the same; the camera walks already skip absent cars via `carInWorld`, so the dial steps past them rather than dispatching at a car iRacing can't focus.
- **Class position still comes from `classPositionFromOrder`.** The grid's own `ClassPosition` field is never read, so class is never sourced independently.
- **The handover at the green can't read as a burst of position changes.** Every event-emitting consumer of the order is `isPreGreen`-gated, and `diffOvertakes` seeds fresh at the first racing tick. Covered by a regression test, which I verified bites by temporarily removing `|| preGreen` from the diff.
- **`resolveLeaderLapTimeS` reports no leader pace pre-green.** Naming a leader there also means reading their `CarIdxLastLapTime`, which on a rolling start is the parade lap — minutes long. Feeding that to the timed-race fuel-coverage estimate under-counts the laps left and *suppresses* fuel warnings, the opposite of that estimate's documented safe direction (#880).
- **`SessionState.Invalid` stays inside the gate.** Session Info's position display already switches to the grid on the same predicate (#647), so the exposure isn't new — and carving it out here would create a second pre-green definition, the exact drift `race-positions.md` exists to prevent.

The second commit is code-review hardening: `Position` is now bounded like `CarIdx` (it becomes the rank, and the dial walks ranks — one absurd value would hang it), and a `QualifyResultEntry` type models the `null` YAML list item the parser already skipped but the translator's grid-slot reads dereferenced unguarded.

## How to test

Manually verified in a rolling AI race (both spectating and driving).

1. Place a Camera Controls instance on a Stream Deck+ dial in **Cycle by Race Position** mode.
2. Start a race and stay on the grid / parade lap. Turning the dial now steps through the field; confirm the order matches iRacing's actual grid (F3), not who happens to be furthest round the lap — the front row is the giveaway, since a distance ranking inverts it.
3. Drivers still in the garage hold their grid number in any position readout, but the dial steps past them.
4. `{{race_ahead.car_number}}` / `{{race_behind.car_number}}` on a Telemetry Display key resolve to the cars you line up between.
5. At the green the dial keeps working as the order rolls to live, and the Race Engineer says nothing about positions at that moment.
6. A standing start behaves the same; a practice session is unchanged (no grid order there by design).

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — n/a, no action, mode, binding, manifest or comms-catalog entry changes; the fix is in the shared translator and SDK layers all three plugins consume

## Summary by CodeRabbit

- **New Features**
  - Race positions now follow the starting grid before the green flag, including formation and parade laps.
  - Live race order takes over automatically once racing begins.
  - Race-position cycling, `race_ahead`, and `race_behind` now use consistent ordering.
  - Added reliable qualifying-grid handling, including gaps and unavailable entries.

- **Bug Fixes**
  - Improved camera cycling fallback when live rankings are unavailable.
  - Prevented duplicate or invalid grid assignments from affecting position estimates.

- **Documentation**
  - Updated documentation with pre-green race-position behavior and starting-grid details.


niklam:
@coderabbitai Good catch — fixed in a2e0415a.

You were right that the assertion was vacuous. `raceSessionWithGrid()` ranks the field `[3, 2, 1]`, and the lap distances I fed it (`[0.1, 0.5, 0.9]`) produced `[3, 2, 1]` as well, so the test held whether or not the race-only gate existed. It proved nothing about a stale qualifying grid staying out of a practice session — which is the entire point of that gate.

Applied your suggestion (lap distances reversed to `[0.9, 0.5, 0.1]`, expectation now `[1, 2, 3]`, plus a comment saying why the inversion is deliberate so nobody "tidies" it back). I verified it now bites rather than just passing: temporarily dropping `isRaceSession` from the gate in `race-order.ts` makes the test fail, and it passes again once restored.

Full verification after the change: `pnpm build` 22/22, 8069 tests, lint and Prettier clean.

PR #1000, opened 2026-08-16
feat(audio-assets): add the Default (Short Calls) Race Engineer voice (#999)

### LINKED ISSUE #999: Race Engineer: add a Short Calls voice with terse, radio-style callouts
## Summary

Add a second Race Engineer voice — id `default-short-calls`, label **Default (Short Calls)** — that says the same things in radio shorthand. It reuses the default voice's ElevenLabs voice id, model and voice settings, so it is the same engineer speaking; only the wording changes. The existing `default` voice ships unchanged and stays the first choice.

| Callout | `default` | Short calls |
|---|---|---|
| Blue flag | "Blue flag. Faster car approaching." | "Blue flag." |
| Checkered (race) | "Checkered flag. That's it for the race, well done." | "Checkered flag." |
| Meatball | "We got the meatball flag. We'll need to pit for repairs." | "Meatball." |
| One pace lap to go | "Still a lap to go before racing. Keep your spacing tidy and work some heat into the tyres." | "One to green." |
| Pits open | "Pits are open." | "Pits open." |
| Refuelling next stop | "We're refuelling at the next pit stop." | "Fuel on." |

## Use case

A driver who already knows what a blue flag means doesn't need the sentence explaining it. Long lines cost attention mid-corner and can still be talking when the next event happens. Short calls keep everything the engineer knows while cutting the time spent parsing it — and anyone who prefers the full engineer simply doesn't switch voices.

## What the work involves

Only the prose lines need rewriting. Most of the ~1,500 entries in `default.voice.json` are value clips — corner names, driver names, position/temperature/speed/lap-time numbers, per-name greetings — already single words that carry over verbatim. The families to shorten are flags, incidents, pit actions/readbacks/status, pit window and limiter, fuel, gap, track conditions, spotter, damage, and the session/race start and end lines.

A short voice may also carry fewer `-NN` variants than the default (one acknowledgment instead of five, say). Variant counts are per-voice and pools are derived from the clips that exist (#664), so that needs no code or pool change.

**No plugin code change.** Voices are auto-discovered from the runtime manifest, so the new voice appears in the **Race Engineer Voice** dropdown (Pit Crew PI → Pit Crew Settings) in all three plugins once the config, clips and manifests are committed.

## Rules to respect

- **Leave `default.voice.json` alone.** Every audio-affecting field feeds the entry hash, so even a `previous_request_ids` edit re-cuts that clip and changes audio that already ships. If the short voice needs a different reference list, it belongs in the short voice's own config.
- **Typo guard** (`src/generate/voice-parity.test.ts`): the short voice may omit callouts and carry fewer variants, but every `<group>/<base>` key it defines must already exist in `default.voice.json`. Anything genuinely new gets added to the default voice first.
- **Scope generation by group** — ElevenLabs is a paid API. Dry-run first (`generate:dry-run --voice default-short-calls --group <group>`), then generate scoped, then `generate:manifest`. Full workflow in `packages/audio-assets/CLAUDE.md`.
- **Dropdown label:** `ird-voice-select` title-cases the raw voice key, so `default-short-calls` renders as "Default-short-calls". Either pick a key that reads well, or teach the component to use the config's `label` field (it exists today but is generator-only).

## Affected artifacts (per `build-and-commit.md`)

- [ ] `@iracedeck/audio-assets` — `configs/default-short-calls.voice.json`, clips under `voice/default-short-calls/`, both manifests rebuilt
- [ ] **Website** — the voice list in `docs/actions/audio-voice/pit-crew.md`, plus a changelog entry (`changelog.mdx`)
- [ ] No plugin, PI, scenario or settings changes — unless the dropdown-label fix is taken, which touches `packages/pi-components/src/components/voice-select.ts`

Requested on Discord: https://discord.com/channels/1477659500851888219/1535054724934664193 by projektdotnet42 (2 ❤️)

projektdotnet:
## Related Issue

Fixes #999

## What changed?

Adds a second Race Engineer voice, `default-short-calls` (**Default (Short Calls)**) — the same engineer, speaking radio shorthand.

- `configs/default-short-calls.voice.json` — 46 groups, 1,474 entries, same ElevenLabs voice id, model and voice settings as `default`.
- 1,474 generated clips under `voice/default-short-calls/`, plus rebuilt `generate.manifest.json` and `manifest.json`.
- 172 lines are reworded shorter ("Blue flag. Faster car approaching." → "Blue flag."; "Still a lap to go before racing. Keep your spacing tidy and work some heat into the tyres." → "One to green."). The remaining ~1,300 are value clips — corner names, driver names, numbers, greetings — carried over verbatim.
- Several families ship fewer variants than the default voice: acknowledgment 5→1, openers 4→1, rolling-start 5→1, damage 3→1, flags 44→30, incidents 22→10, opponent-pit 13→6, pit-limiter 7→4, pit-window 10→2.

No code changes: the voice is auto-discovered from the manifest and shows up in the **Race Engineer Voice** dropdown (Pit Crew PI → Pit Crew Settings). The `default` voice stays the fallback.

### Also touches the default voice

`configs/default.voice.json` drops `pit-actions/acknowledgment-03` from `previous_request_ids` on 24 pit-action entries (fuel, tires, compounds, windshield, fast repair). Reference strings feed the entry hash, so those 24 `voice/default/pit-actions/*.mp3` clips were re-cut and are included here — shipped default-voice audio changes in this PR. The short voice can keep its own two-entry reference list without the default voice changing, so this can be reverted.

## How to test

1. `pnpm install && pnpm build`
2. Open any **Pit Crew** button's Property Inspector → **Pit Crew Settings** → **Race Engineer Voice**, select the short-calls voice, and press **Test** next to Race Engineer Volume.
3. Trigger callouts in a session (flags, pit actions, spotter) and compare against the default voice. The scenario harness (`pnpm --filter @iracedeck/scenario-harness dev`) also works, but it seeds the first voice alphabetically (`default`) — set `raceEngineerVoice` to the short voice to audition it there.
4. `pnpm test` — `voice-parity` currently fails: the short voice defines `openers/ok`, which `default.voice.json` doesn't have, and the typo guard rejects bases the default voice doesn't know (such a base is in no pool, so it would never play). Either add `ok` to the default voice's `openers` group or reuse an existing opener.

## Checklist

- [ ] New code has unit tests — n/a, assets only
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — n/a, voices are manifest-derived in all three plugins
- [ ] No unrelated changes included — see the default-voice note above

## Summary by CodeRabbit

- **New Features**
  - Added an expanded set of concise voice announcements for acknowledgments, corner names, flags, pit status, race and session updates, spotter calls, and measurement units.
  - Added short-call audio coverage for more in-game communication scenarios.

- **Bug Fixes**
  - Updated pit-action voice selection to avoid using a specific acknowledgment clip for fuel, tire, compound, windshield, and fast-repair messages, improving consistency with available audio.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "iracedeck",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "private": true,
   "packageManager": "pnpm@10.0.0",
   "engines": {
--- a/packages/audio-assets/package.json
+++ b/packages/audio-assets/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/audio-assets",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Shared audio assets (MP3) for iRaceDeck plugins",
   "type": "module",
   "exports": {
--- a/packages/audio-native/package.json
+++ b/packages/audio-native/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/audio-native",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Native Node.js addon for miniaudio-based multi-channel audio mixer",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/packages/audio-scenarios/package.json
+++ b/packages/audio-scenarios/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/audio-scenarios",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Data-driven audio scenario DSL and interpreter. Subscribes to @iracedeck/event-bus and drives @iracedeck/audio-service.",
   "type": "module",
   "main": "dist/index.js",
--- a/packages/audio-service/package.json
+++ b/packages/audio-service/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/audio-service",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Multi-channel audio mixer singleton over @iracedeck/audio-native",
   "type": "module",
   "main": "dist/index.js",
--- a/packages/callout-script/package.json
+++ b/packages/callout-script/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/callout-script",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "The JSON grammar for Race Engineer voice-pack callout scripts: types, Zod schema, parser and reference collector.",
   "type": "module",
   "main": "dist/index.js",
--- a/packages/deck-adapter-elgato/package.json
+++ b/packages/deck-adapter-elgato/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/deck-adapter-elgato",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Elgato Stream Deck adapter for @iracedeck/deck-core",
   "type": "module",
   "main": "dist/index.js",
--- a/packages/deck-adapter-mirabox/package.json
+++ b/packages/deck-adapter-mirabox/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/deck-adapter-mirabox",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Mirabox adapter for @iracedeck/deck-core",
   "type": "module",
   "main": "dist/index.js",
--- a/packages/deck-adapter-ulanzi/package.json
+++ b/packages/deck-adapter-ulanzi/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@iracedeck/deck-adapter-ulanzi",
-  "version": "3.2.0",
+  "version": "3.3.0-dev.0",
   "description": "Ulanzi Deck adapter for @iracedeck/deck-core",
   "type": "module",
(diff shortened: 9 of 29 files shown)


packages/audio-assets/configs/default-short-calls.voice.json
```diff
@@ -0,0 +1,8367 @@
...
+        "text": "Copy."
+      },
+      {
+        "name": "acknowledgment-04",
+        "text": "Copy."
+      },
+      {
+        "name": "acknowledgment-05",
+        "text": "Copy."
+      }
```
niklam, line 17: You don't need to have any variations after `-01`. Variations defines a group from which the callout is randomly chosen from. So in this case we're randomly choosing from five "Copy" calls.

The preferred way to have the full sentences — even if one word — is to have them as `"Copy."`. Everything affects how text-to-speech engine treats the strings. This doesn't obviously apply to sentence parts such as `"seconds"` below.

packages/audio-assets/configs/default-short-calls.voice.json
```diff
@@ -0,0 +1,8367 @@
...
+      },
+      {
+        "name": "blue-01",
+        "text": "Blue flag.",
+        "seed": 3
+      },
+      {
+        "name": "blue-02",
+        "text": "Blue flag.",
+        "seed": 2
```
niklam: As no audio files has been generated, every single `"seed"` can be removed. Seed defines the variation of the call. So once you create the calls with Elevenlabs (I will help you with that), you're going to have to listen to all of them, and if some are "broken", increase seed by one, re-create, test, and repeat until the callout is good.

niklam (changes requested):
Just a couple of comments!

PR #1004, opened 2026-08-18
improve(stream-deck): drop the macOS support claim from the plugin manifest (#994)

### LINKED ISSUE #994: Drop the macOS support claim from the Stream Deck manifest
### What would you like?

Drop the macOS support claim from the Stream Deck plugin manifest.

`packages/iracing-plugin-stream-deck/com.iracedeck.sd.core.sdPlugin/manifest.json` currently declares:

```json
"OS": [
  { "Platform": "windows", "MinimumVersion": "10" },
  { "Platform": "mac", "MinimumVersion": "15" }
]
```

The `mac` entry should go, leaving Windows only — which is what the Mirabox and Ulanzi manifests already declare.

### Use case

iRacing is Windows-only, so an iRacing plugin cannot do anything useful on macOS. The `mac` entry exists for a historical reason that no longer applies: it made it possible to develop the plugin on a MacBook. That development has stopped.

Meanwhile the entry has a live consequence — the manifest `OS` array is what the Elgato Marketplace uses to decide who is offered the plugin. A Mac user can currently be shown, and can install, a plugin whose every action depends on a simulator that does not run on their machine. It advertises support that does not exist.

This is small and unrelated to any feature work; it just needs doing.

### Area

Stream Deck plugin

### Alternatives considered

- **Leave it.** Nothing breaks for existing users, and the audience affected is small. But the claim stays wrong, and it is a one-line fix.
- **Keep it and add a runtime warning on macOS.** More work to build and maintain than simply not advertising a platform we do not support.

### Additional context

**The non-Windows native mocks stay.** `@iracedeck/iracing-native` and `@iracedeck/audio-native` both fall back to mock implementations off Windows, and the `cross-platform-development` skill documents that workflow. Those are development tooling, not a support claim — removing the manifest entry does not mean removing them, and they should be left alone so building and testing on a Mac remains possible.

**Also worth checking while in here:**

- The Elgato Marketplace listing itself, which lives outside this repo and may carry its own platform metadata or copy.
- `Software.MinimumVersion` and the `OS` entries on all three manifests, for consistency while the file is open.
- The website and `README.md` were checked and contain **no** macOS claims today, so nothing is needed there.

Note that dropping a declared platform may need care on the Marketplace side — worth confirming how Elgato handles an update that narrows the `OS` array for users who already have it installed, rather than assuming it is silent.

niklam:
## Related Issue

Fixes #994

## What changed?

The Elgato manifest's `OS` array no longer declares `mac`. All three plugin manifests now declare Windows 10 and nothing else, which is what the Mirabox and Ulanzi manifests already said. The array is what the Elgato Marketplace reads to decide who is offered the plugin, so until now a Mac user could be shown — and could install — a plugin whose every control drives a simulator that does not run on their machine.

`scripts/manifest-platform.test.mjs` guards the invariant mechanically, following the existing `manifest-actions-order.test.mjs` pattern: every discovered plugin manifest must declare exactly `["windows"]`, and no action may re-introduce the claim through the per-action `OS` list the Elgato schema allows. Manifests are discovered via `allPluginManifestRelPaths`, so a future deck ecosystem is covered without editing a parallel hardcoded list.

**Consistency check requested by the issue:** the three `Software` minimum versions are deliberately left alone. They name three different host applications — Stream Deck `7.1`, Mirabox `3.10.188.226`, UlanziStudio `2.1.0` (under Ulanzi's own `MinVersion` key) — and are not comparable to one another. The Stream Deck `7.1` already matches what the installation page tells users.

**The non-Windows native mocks are untouched**, per the issue. `plugin-structure.md` gains a Supported Operating Systems section and the `cross-platform-development` skill now states outright that the mocks are development tooling rather than a support claim, so the manifest entry does not get restored on their account.

Website: the installation page's three Requirements lists and the downloads page now state the Windows requirement, which none of them did before — a Mac visitor was previously told nothing either way.

**Open question for the release, not for this PR:** Elgato does not document how the Marketplace treats an update that narrows the `OS` array for users who already have the plugin installed. The manifest reference defines `OS` only as "Collection of operating systems, and their minimum required versions, that the plugin supports" and is silent on the rest. The changelog entry therefore claims only what is certain — that the plugin is offered and installable on Windows only from this version — and makes no promise about existing Mac installs. The Marketplace listing's own platform metadata lives outside this repo and is worth a manual check at release.

## How to test

1. `pnpm install && pnpm build && pnpm test` — 284 test files, 8344 tests.
2. `npx vitest run scripts/manifest-platform.test.mjs` — 7 tests. Temporarily re-add a `mac` entry to any manifest, or an `"OS": ["mac"]` to any action, and confirm the run fails.
3. `streamdeck validate ./packages/iracing-plugin-stream-deck/com.iracedeck.sd.core.sdPlugin` — Validation successful.
4. Link the plugin and start the Stream Deck app: the iRaceDeck category and all its actions appear as before (verified manually on Windows).
5. `pnpm --filter @iracedeck/website build` — the Windows requirement renders on `/docs/getting-started/installation/` (three times) and `/downloads/`.

## Checklist

- [x] New code has unit tests

https://claude.ai/code/session_013DfEgcdKE8JrdCP1ReYc7X

## Summary by CodeRabbit

* **Documentation**
  * Clarified that iRaceDeck supports Windows only.
  * Added a Windows 10 or newer requirement to installation and download guidance.
  * Updated the changelog to note the removal of macOS support and documented the platform requirements.

* **Bug Fixes**
  * Updated plugin platform declarations to prevent unsupported macOS installations.

* **Tests**
  * Added validation ensuring all plugin and action declarations target Windows 10 or newer.

### DIFF
--- a/.claude/rules/plugin-structure.md
+++ b/.claude/rules/plugin-structure.md
@@ -134,6 +134,18 @@ Every plugin artifact must ship the project `LICENSE` and the aggregated `THIRD-
 
 When a shipped third-party dependency or vendored component is added or removed, update the repo-root `THIRD-PARTY-LICENSES.md` in the same change. `scripts/third-party-licenses.test.mjs` guards the wiring: every non-workspace rollup `external` must have an entry in the file, every plugin config must contain the copy step, and no `.sdignore` pattern may exclude the two files from the packed plugin.
 
+### Supported Operating Systems
+
+Every plugin manifest declares **Windows and nothing else** (issue #994):
+
+```json
+{
+  "OS": [{ "Platform": "windows", "MinimumVersion": "10" }]
+}
+```
+
+iRaceDeck exists to control iRacing, which has no macOS build, and the keyboard/window/audio addons are Windows-native. The `@iracedeck/iracing-native` / `@iracedeck/audio-native` mocks exist so the repo can be installed, built, and tested on a Mac or Linux machine (see the `cross-platform-development` skill) — they are **not** a reason to add a `mac` entry, which would only offer host users an install that can do nothing. `scripts/manifest-platform.test.mjs` guards this across every discovered plugin manifest (plugin-level `OS` and per-action `OS` alike), so a new plugin folder is covered automatically.
+
 ### Application Monitoring
 
 To enable app monitoring (for features like conditional reconnection that pauses when iRacing isn't running), add to manifest.json:
--- a/.claude/skills/cross-platform-development/SKILL.md
+++ b/.claude/skills/cross-platform-development/SKILL.md
@@ -7,7 +7,7 @@ description: Use when working with native dependencies, handling platform differ
 
 ## Why the mock exists
 
-**The mock is development tooling, not a support claim.** iRaceDeck ships Windows-only — all three plugin manifests declare `windows` and nothing else (#994) — because every action depends on iRacing, which has no macOS build. The mock exists so the repo can still be installed, built, and tested on a Mac or Linux machine; it is not there to make the plugin usable off Windows, and it is not a reason to re-add a `mac` entry to a manifest.
+**The mock is development tooling, not a support claim.** iRaceDeck ships Windows-only — all three plugin manifests declare `windows` and nothing else (#994) — because iRaceDeck exists to control iRacing, which has no macOS build. The mock exists so the repo can still be installed, built, and tested on a Mac or Linux machine; it is not there to make the plugin usable off Windows, and it is not a reason to re-add a `mac` entry to a manifest. `scripts/manifest-platform.test.mjs` fails the build if one comes back.
 
 `@iracedeck/iracing-native` wraps the iRacing SDK (Windows-only C++ N-API addon). Without the mock layer, `pnpm install`, `pnpm build`, and `pnpm test` all fail on macOS/Linux because:
 - `node-gyp` can't compile the Windows-specific C++ code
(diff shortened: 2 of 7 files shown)


niklam:
@coderabbitai Thanks — the nitpick was valid and is fixed in 030507ea.

You were right that the guard checked only `Platform`, so a manifest that dropped `MinimumVersion` or moved it off `"10"` passed silently. That matters a little more than a generic tightening here: issue #994 explicitly asked to check the `OS` entries and minimum versions across all three manifests for consistency, so the version belongs inside the same guard rather than resting on the PR description.

I implemented it slightly differently from the suggested diff — asserting the whole array in one comparison instead of indexing `OS[0]`:

```js
expect(readManifest(manifestRelPath).OS).toEqual([{ Platform: "windows", MinimumVersion: "10" }]);
```

Reasoning: `OS[0].MinimumVersion` is only safe because the preceding assertion happens to have proved there is exactly one entry, which couples the two lines in a way that is easy to break later. One `toEqual` states the whole invariant, needs no index, and also rejects a stray extra entry that carries the right platform. The test name is now "declares Windows 10 and nothing else".

Verified both directions rather than just the happy path:

- Positive: `npx vitest run scripts/manifest-platform.test.mjs` → 7 passed; full suite 284 files / 8344 tests.
- Negative: temporarily set the Ulanzi manifest's `MinimumVersion` to `"11"` → the guard fails on that manifest as intended; manifest restored afterwards.
- Also re-confirmed the original guards still bite: re-adding a `mac` entry at plugin level and an `"OS": ["mac"]` on an action each fail their respective assertion.

One deliberate consequence worth naming: pinning `"10"` means a genuine future move to a Windows 11 floor has to edit this test. That is intended — a platform floor should not shift without someone consciously deciding to narrow who is offered the plugin.

Issue #1005, opened 2026-08-18
Recover gracefully when the settings window can't open


niklam:
### What would you like?

When the settings window cannot be opened, say so in the Property Inspector instead of leaving the user with a button that does nothing — and decide what, if anything, they can still reach.

Two things:

1. **A warning banner** (`setWarning`) whenever the settings server fails to start or the window fails to open, cleared as soon as one succeeds. Every PI already renders `_warnings` through the auto-injected `ird-warnings` component, so this needs no per-template markup.
2. **A decision on a fallback** for reaching plugin-global settings while the window is unreachable. Options and a recommendation are below — this half deserves an explicit call rather than being folded in.

### Use case

Both failure paths exist today and are **log-only**, identically in all three plugins:

| Path | Stream Deck | Mirabox | Ulanzi |
|---|---|---|---|
| Server failed to bind — `ensureStarted()` rejects | `plugin.ts:984` | `plugin.ts:943` | `plugin.ts:946` |
| Window failed to open — `open()` rejects | `plugin.ts:1156` | `plugin.ts:1081` | `plugin.ts:1087` |

The second only fires when the chromeless app window **and** the `openUrl` tab fallback both failed, so it is genuinely rare. The first is the more plausible one: an aggressive security suite or a port conflict stops the loopback listener, and nothing about that reaches the user — they press *Open iRaceDeck Settings* and nothing happens.

**#1003 is what makes this worth fixing.** It did not make failure more likely; it raised the cost. Before it, a settings window that would not open was an annoyance, because every plugin-global setting was still editable in each action's Property Inspector. Since #1003 the window is the only route to appearance defaults, the Race Engineer voice and callouts, SimHub, dual-press, debug logging and the rest. Same probability, much larger blast radius — and no signal at all that something is wrong.

### Area

Stream Deck plugin

### Alternatives considered

- **Leave it.** The realistic failure rate is low and the logs already record it. Rejected because the one thing the user gets today is a dead button, and a dead button reads as a broken plugin rather than a blocked port.
- **Re-embed the global settings in every Property Inspector, hidden, revealed on failure.** This is the obvious "fallback UI" answer and it is the wrong one — see the analysis below. It reinstates the ~35× duplication and the two-editors problem that #1003 removed, permanently, to cover a rare case.
[...]

Issue #1003, opened 2026-08-18
Remove the duplicated plugin-global settings from every action Property Inspector


niklam:
### What would you like?

Remove the plugin-global settings from all 35 action Property Inspectors, now that the dedicated settings window (#992, #993) owns them. This is the second half of #992 — that issue's stated end state already says "the other global accordions (title / color / border / graphic defaults, flag flash, common settings) move to the window", but the window was built and the PIs were left intact. Today both surfaces render the same settings, backed by the same plugin-owned store.

The one global surface that **stays** in the PI is the `Related Key Bindings` accordion. Each PI renders only its own category out of 238 bindings, which makes it genuinely contextual to the action being configured — and `ird-binding-status` depends on it (see the trap below). Everything else goes.

Per-PI end state, replacing the current `Global Settings` section:

```text
Action Settings
  …action-specific settings…
  Title / Color / Border / Graphic Overrides   ← per-action, unchanged
  Common Settings (flags overlay)              ← per-action, unchanged

Key Bindings                    [Open iRaceDeck Settings]
  Related Key Bindings          ← stays (25 PIs have one; 10 do not)
```

`pit-crew.ejs` is in scope too, and needs looking at specifically: it renders three plugin-global accordions **above** the Global Settings header, inside `Action Settings`, so a mechanical clean-out walks straight past them. See "Pit Crew" below.

### Use case

Plugin-global settings are now editable in two places at once. Since #993 both surfaces write through the same plugin-owned store, so the PI copies are pure redundancy rather than a fallback — two UIs to keep in sync, two places for a user to look, and a settings panel that is still mostly not about the action you opened it from.

It also finishes the job #992 set out to do. Shipping the settings window while leaving every PI unchanged is the worst of both worlds: users get a new window *and* still scroll past six global accordions on every key.

### Area

Stream Deck plugin

### Alternatives considered

- **Remove the key bindings too, for a fully clean PI.** Rejected for this issue, worth revisiting later. It needs `ird-binding-status` rewired off DOM-scraping and onto the #993 settings bridge, plus a deep link into the settings window's Key Bindings table filtered to one binding. That is a real feature, not a clean-out, and it would hold this change hostage.
- **Leave the PI sections in place as a convenience shortcut.** Rejected. Duplicated UI is what motivated #992; keeping it halves the benefit and doubles the maintenance. Discoverability is already handled — every PI keeps the `Open iRaceDeck Settings` button.
[...]

PR #1025, opened 2026-08-23
improve(pi-components): move the iRaceDeck Settings button under each action's own settings (#1024)

### LINKED ISSUE #1024: Move the iRaceDeck Settings button under the action's own settings, and make it discreet
### What would you like?

Move the *Open iRaceDeck Settings* button from the bottom of every action Property Inspector to directly under that action's **own** settings — above `Title Overrides` — and restyle it so it stops being the loudest thing on the panel: a cog glyph in front, the shorter label **iRaceDeck Settings**, auto width, smaller text, brand red kept.

It **moves**; it does not appear twice. `key-bindings-section.ejs` stops rendering it, and the docs link becomes that section's only tail.

```text
Title: [                          ]   <- deck software's own field, untouchable

ACTION SETTINGS
  Mode: [ Race Engineer Toggle  v ]
  …action-specific settings…
  ─────────────────────────────────
         [ ⚙ iRaceDeck Settings ]     <- here, discreet
▸ TITLE OVERRIDES
▸ COLOR OVERRIDES
▸ BORDER OVERRIDES
▸ GRAPHIC OVERRIDES
▸ COMMON SETTINGS

KEY BINDINGS
  ▸ Related Key Bindings
  See documentation for this action    <- now the only tail
  Version: 3.0.0-dev.0 - Downloads
```

### Use case

Requested on Discord by **projektdotnet42**: *"Suggestion, if possible, move this button to the top."*

Since #1003 the settings window owns every plugin-wide setting, which makes this button the most-used control on a Property Inspector — and it is the last thing on the page. Reaching it means scrolling past four override accordions and the Key Bindings section on a panel most users opened for one dropdown. Above the overrides it is visible without scrolling on nearly every action.

The reason it can't simply go to the top is the panel's first row: the **Title** field belongs to the deck software and nothing can be rendered above it. Under the action's own settings is as high as a plugin control can sit — and it is also the honest boundary: everything below that line (title, colour, border, graphic, flags) is *this key's* appearance, everything behind the button is *every* key's.

The trade is prominence. As a section footer, a full-width brand-red button read as a closing call to action. The same red directly under the Mode row competes with the settings the panel is actually about, so it shrinks: cog + shorter label, auto width, smaller text.

### Area

Stream Deck plugin

### Alternatives considered

- **Both places — new spot plus the existing footer.** Rejected. Two buttons for one destination is the duplication #1003 spent a whole issue removing, in miniature.
- **A text link, styled like the docs link.** Rejected as too quiet: opening a separate window deserves a button's affordance, and the docs link directly below would then be two adjacent links to different kinds of destination.
- **Top of the panel, above `Action Settings`.** Impossible above the host's Title field; directly under it would put a plugin-wide control ahead of the settings the key was opened for.
- **Inject it from the top of `title-overrides.ejs`** — a tempting one-file change that reaches all 35 PIs. Rejected, and it is a trap: see the dial note below.

#### Where exactly it goes — and the dial trap

All 35 action PIs include `title-overrides` immediately after their own settings, so the boundary is well-defined. But on the **16 dial-capable PIs** that include sits inside `<div id="keypad-appearance">`, which the PI hides on the dial surface (a dial has no key icon):

```text
<div id="keypad-settings">…</div>     <- hidden on the dial surface
<div id="dial-settings">…</div>       <- hidden on the keypad surface
        ^ the button belongs HERE, between the two wrappers and #keypad-appearance
<div id="keypad-appearance">
  title-overrides / color / border / graphic / common-settings
</div>                                <- hidden on the dial surface
```

Anything anchored to `title-overrides` — including injecting the button from inside that partial — vanishes on the dial surface of `audio-controls`, `black-box-selector`, `camera-editor-adjustments`, `camera-focus`, `cockpit-misc`, `force-feedback`, `fuel-service`, `setup-aero`, `setup-brakes`, `setup-chassis`, `setup-engine`, `setup-fuel`, `setup-hybrid`, `setup-traction`, `splits-delta-cycle`, `view-adjustment`. So the include goes into each action template explicitly, **outside every surface-conditional wrapper** — 35 edits, one of them per file, not one clever include.

`.ird-section-footer` is reused as-is: its `border-top` draws exactly the rule the mockup shows above the button, and in the new position it genuinely closes the Action Settings block. Only its CSS comment (which cites #1003 and says "below the section's own content") needs rewording.

#### Styling

- `defineSendToPluginButton` gains an optional `icon` (inline SVG, `currentColor`) rendered before the label. Opt-in, so `ird-open-folder` (#993) in the settings window's Diagnostics tab is untouched.
- Each component already injects its own `${tag} button` block, so shrinking `ird-open-settings` cannot leak into `ird-open-folder`.
- `ird-open-settings`: `defaultLabel` → `"iRaceDeck Settings"`, cog icon, smaller font/padding, `width: auto` (the `.ird-open-settings` wrapper already un-stretches it; drop `width: 100%; min-width: 200px` from the tag rule rather than relying on the wrapper).
- Prefer an inline SVG cog over the `⚙` character — U+2699 can render as a colour emoji depending on the font fallback in the host webview.

#### The failure banner travels with the button (#1005, #1014)

`open-settings.ejs` renders `<ird-warnings only="settings-window-open">` immediately above the button, and the button's "unusable" note sits with it. Moving the include moves all three together, and `head-common.ejs`'s matching `except` keeps the top strip free of that id regardless of position — so nothing breaks. What changes is the *rationale*: #1005 argued the banner must be beside the button because the button is "a full scroll from the top strip". After this it is a few centimetres from it. Keep the banner with the button anyway (it explains that specific control), but update the reasoning in `.claude/rules/settings-window.md` so the next reader isn't told a fact that is no longer true.

#### Label rename ripples

`Open iRaceDeck Settings` → `iRaceDeck Settings` appears as a literal in prose across the repo. Every one of these also states the old *position* ("at the bottom of any action's Property Inspector"), so both halves need the same pass:

| File | What to fix |
|---|---|
| `.claude/rules/settings-window.md` | button row (label, placement, #1005 rationale) |
| `.claude/rules/pi-templates.md` | `section-header` / `key-bindings-section` / `open-settings` entries |
| `.claude/rules/stream-deck-actions.md` | `ird-open-settings` component entry |
| `.claude/rules/global-settings.md` | key-bindings-section description + the `_warnings` placement paragraph |
| `.claude/rules/logging.md` | "point the user at the … button" |
| `docs/architecture/pi-template-system.md` | superseded-note wording |
| `packages/website/…/getting-started/settings.md` | how-to-open instructions (×2) + the failure section |
| `packages/website/…/getting-started/troubleshooting.md` | Diagnostics how-to |
| `packages/website/…/actions/audio-voice/pit-crew.md` | Race Engineer settings pointer |
| `packages/website/…/features/stream-deck-profiles.md` | profile-install step 1 |
| `packages/website/…/changelog.mdx` | the 3.0 entry describes the old position — **edit that line**, don't add a new one (3.0 is unreleased) |

#### Tests to update

`key-bindings-section-partial.test.ts` (asserts the button is part of that section), `open-settings.test.ts`, `send-to-plugin-button.test.ts` (new `icon` option), `warnings.test.ts`.

#### Out of scope

`settings.ejs`, the plugin-level fallback PI, keeps its button under its own header via `section-header.ejs`'s `openSettings` slot. It picks up the restyle for free, which is correct — that page is nothing *but* the way through.

#### Target branch

`release/3.0` — the settings window and this button exist only there.

niklam:
## Related Issue

Fixes #1024

## What changed?

The *iRaceDeck Settings* button moves from the bottom of every action Property Inspector to directly under that action's **own** settings, above Title Overrides — and gets dressed for the new spot: a cog, the shorter label **iRaceDeck Settings**, and the button factory's new `compact` size.

Since #1003 the settings window owns every plugin-wide setting, which makes this button the most-used control on a Property Inspector — and it was the last thing on the page, a scroll past four override accordions and the Key Bindings section away. Under the action's own settings is as high as a plugin control can go: the panel's first row is the deck software's own Title field. It also lands on an honest boundary — everything below it is one key's appearance, everything behind it is every key's.

It **moves**; it does not appear twice.

```text
ACTION SETTINGS
  Mode: [ Race Engineer Toggle  v ]
  ─────────────────────────────────
         [ (cog) iRaceDeck Settings ]   <- here
TITLE / COLOR / BORDER / GRAPHIC OVERRIDES
COMMON SETTINGS

KEY BINDINGS
  Related Key Bindings
  See documentation for this action     <- now the section's only tail
```

Implementation notes:

- New partial `action-settings-footer.ejs` owns the placement rule and is included by all 35 action templates; `key-bindings-section.ejs` no longer renders the button.
- **The dial trap:** on the 16 dial-capable PIs `title-overrides` sits inside `<div id="keypad-appearance">`, which the PI hides on the dial surface. Anchoring the button to it — including the tempting one-line version, rendering it from `title-overrides.ejs` — would have stripped every dial user of the only route to the settings window. `action-settings-footer-partial.test.ts` checks the placement in all 35 templates, in every surface-conditional wrapper.
- `defineSendToPluginButton` gains optional `icon` and `defaultSize`, plus a per-element `size` attribute (the way `label` overrides `defaultLabel`). `ird-open-settings` defaults to `compact`; `settings.ejs` — the page that is nothing but this button — asks for `standard`, so that page is unchanged. `ird-open-folder` is untouched.
- The open-failure banner (#1005) travels with the button, as it should: a warning about one control belongs beside that control. Its original *distance* argument is spent, and the comments that stated it have been corrected rather than left to mislead.
- Docs, rules and the in-development 3.0 changelog line updated for both halves of the change (new label, new position), including the website's Settings page, which now describes where the button is and the key-vs-plugin boundary it marks.

A `/code-review xhigh` pass ran over the branch; its findings are fixed in b89c702, except the #1005 two-banner overlap, which is that issue's behaviour rather than this move's.

## How to test

1. Select any iRaceDeck key. The cog button sits directly under the action's own settings, above `TITLE OVERRIDES`, with a divider above it — and nothing at the bottom but Key Bindings and the docs link.
2. Press it: the Settings window opens as before.
3. On a Stream Deck+ dial (Fuel Service, Setup Chassis, Camera Controls…), switch to the dial surface — the button is in the same place there. This is the regression the placement test guards.
4. An action with no bindings (Pit Crew, Tire Service): button present, "This action has no key bindings." still at the bottom.
5. Fuel Service: switch modes — the Key Bindings section still hides and reveals per mode, and the button never goes with it.
6. The plugin-level fallback panel (`ui/settings.html`) still shows the full-width button under its heading.

Verified locally on Stream Deck hardware, plus `pnpm build` (22/22), `pnpm lint`, `pnpm format`, `pnpm test` (309 files / 8816 tests) and `pnpm --filter @iracedeck/website build`.

## Checklist

- [x] New code has unit tests

## Summary by CodeRabbit

* **New Features**
  * Added a compact, cog-styled **iRaceDeck Settings** button beneath each action’s settings.
  * Reordered Property Inspector content so settings appear before the button, title overrides, and key bindings.
  * Added clearer settings-window failure messaging positioned alongside the settings control.

* **Documentation**
  * Updated setup, troubleshooting, architecture, and action guidance to reflect the renamed button and revised placement.

* **Tests**
  * Added coverage validating button placement, appearance, accessibility, and consistent rendering across actions.


niklam:
@coderabbitai review

PR #1058, opened 2026-08-30
docs(settings): pin the uncorrelated migration read as deliberate (#1053)

### LINKED ISSUE #1053: The migration read accepts any global-settings payload, including a Property Inspector's own save echo
### Problem

While the one-time settings migration is waiting for the deck host, `initGlobalSettings` ingests **any** `didReceiveGlobalSettings` that arrives as the host's answer. The guard is only:

```ts
if (storeReady || migrationDone || !migrationRequested) return;
```

There is no correlation between the read that went out and the payload that comes back — deck-core has no request/response pairing to correlate with. So during that window (`MIGRATION_TIMEOUT_MS`, 10 s) a payload from an entirely different source is accepted as the migration result: it sets `migrationDone`, becomes the cache, is persisted, and decides the store's source.

The realistic source is a **Property Inspector's own save echo**. A PI that never switched to the loopback settings channel talks to the deck host directly, and on hosts that echo a write back, that echo arrives here looking exactly like an answer. The comment above the guard already names this — "only fallback-path PI saves … and, on hosts that echo, the plugin's own mirror write arrive here now, and none of them are ingested" — which is true only because nothing normally re-opens the window after startup.

### Why it is worth a record rather than a PR paragraph

**This is not specific to #1047.** The same window exists today for every fresh install's first migration, and has since #993. #1047 only makes it reachable a second time: once per plugin version, on a store that was previously given up on. That install is also the worst case for it, because a store whose mirror is suppressed never publishes `_settingsChannel`, so *every* PI on it is on the fallback path by construction.

So the hazard outlives #1047 and should not be tracked inside it.

### Mitigations already in place (why this is not urgent)

Three things landed in #1047 that reduce it substantially, none of which close it:

1. **The window opens once per version, not three times.** The retry is a single attempt that records its own give-up either way, rather than restarting the three-start countdown.
2. **An empty or `null` payload retires nothing.** Those are coerced to `{}` and no longer count as a positive read, so the most likely malformed echo cannot clear the guard or re-open the host mirror.
3. **On the retry the file wins outright.** The merge is `{ ...host, ...file }` rather than `mergeMigration`, so an errant echo can only *add* keys the settings file has never held — it cannot overwrite anything the user has. Before that change a host-shaped payload could beat file values sitting at a schema default.

Worst case after those: some keys the file never had get filled from a PI's snapshot of the host copy. Bad, but bounded, and not the settings-destroying shape #1041 was about.

### What closing it actually needs

Request/response correlation the plugin does not currently have. Options, roughly in order of appeal:

- **Correlate on the wire.** Elgato's protocol has no request id for `getGlobalSettings`, so this would mean an adapter-level convention (e.g. the adapter tags the read it issued and only forwards the first reply that follows it, dropping unsolicited pushes while a read is outstanding). Per-adapter work, three implementations.
- **Ignore payloads that arrive while a PI is known to be open.** Cheap, but the plugin does not reliably know, and on Ulanzi a PI's presence is only inferred from a `propertyInspectorDidAppear` marker.
- **Shrink the window.** Not a fix — it changes the odds, not the failure — and the timeout is load-bearing for slow hosts.
- **Accept and document.** Defensible given the mitigations above; this issue then exists to stop the hazard being rediscovered from scratch.

No recommendation offered here deliberately — the trade needs a look at all three adapters, and the first option is the only real fix among them.

### Affected artifacts

- `packages/deck-core/src/global-settings.ts` — the acceptance guard and whatever correlation is chosen.
- The three adapters, if correlation lands at that layer.
- `.claude/rules/global-settings.md` — the migration section states the "asked for" rule, which is what would change.
- Tests: a payload arriving mid-window from a source that is not the read.

### Context

Found by the max-level review of #1047 (finding 3), verified against the code, and deliberately declined there as unfixable within that change's scope. The three mitigations above were applied in the same PR.

niklam:
## Related Issue

Fixes #1053

## What changed?

Nothing behavioural. This makes an accepted hazard legible where the code lives, and corrects three statements about it that were wrong.

While the one-time settings migration window is open, `initGlobalSettings` takes the **first** `didReceiveGlobalSettings` to arrive as the answer, whatever sent it — `migrationRequested` is a boolean, not a correlation. #1053 decided to accept that rather than build correlation. The design record is the spec on master (`docs/superpowers/specs/2026-08-30-issue-1053-migration-read-payload-correlation.md`); this PR is the part that has to live next to the code.

- **The comment above the subscription** said "only when we asked" without saying what asking buys. It now says, and points at the spec and the rules file rather than restating the argument — the same explanation had already been written three times and the copies had already drifted.
- **The ignore branch's** "none of them are ingested" read as a claim about the listener. It describes that branch: the window is shut, or — on an ordinary start with a settings file, the common case — was never opened at all.
- **`mergeMigration`'s doc comment** was stranded above `pendingMigrationStarts`, which has its own doc block, so it attached to nothing and the function had none. Moved onto it, and extended to say what the review made clear: this merge bounds nothing.
- **`.claude/rules/global-settings.md`** gets the same correction plus the per-host reason correlating is not worth doing.

### Two claims here were wrong before review, and one of them matters

The damage bound was generalised from one path to all three. `{ ...host, ...file }` is only the give-up retry; the fresh and `_migrationPending` paths go through `mergeMigration`, which bounds nothing — with no file the payload becomes the whole cache, and a defaults-born file loses every key still at its schema default. What makes the fresh path acceptable is that there is nothing yet to lose, which was never stated.

"Mirabox's read carries nothing that comes back" was an unmeasured negative asserted as protocol fact, one line from a row labelled **Measured** citing a live sweep. `VSDEvent` declares a `context?` the request already sets, and our adapter narrows to `payload.settings` before deck-core sees anything — structurally the same "exists but discarded" case as Elgato. It is untested, not absent. The spec was amended on master for the same reason (`e6668cfb`).

### No new tests, and two removed

Two regression tests were written for this and then deleted. Neither could fail when the behaviour it named was reverted: the retry test set `driverName` against a schema default of `""`, so `mergeMigration` preserves it exactly as the retry merge does — swap the merge back and it still passes, while its comment claimed to guard that swap. The other reproduced two existing tests minus two of their assertions.

The behaviour was already pinned; what was missing was that the pin looked accidental. A test that cannot fail is worse than none — it turns an absent guard into an apparent one and invites someone to delete the real guard as redundant. So the rationale moved onto the tests that do discriminate, and the test file's net change is comments only.

### One review finding declined

The review asked for the other half of the retry bound to be pinned — that `{ ...raw, ...migrationBase }` can never drop a file key. Declined: that is object-spread semantics, not behaviour, so a test would pin the language rather than the code. Every other finding was applied.

## How to test

Nothing runnable — no behaviour change, no new tests, no settings-key or schema change. The wording was reviewed and approved directly instead, which is the substitute `issue-workflow.md` prescribes when there is nothing to exercise.

Green on the committed state: build 22/22 (deck-core a cache miss, so `tsc` ran), format clean, lint clean, full suite 317 files / 9026 tests, working tree clean. 9026 is the baseline — the two added tests are gone again.

## Checklist

- [x] New code has unit tests — n/a, no new code; see "No new tests, and two removed" above for why none were added
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — n/a, nothing plugin-facing

## Summary by CodeRabbit

* **Documentation**
  * Clarified migration behavior when host responses arrive unexpectedly or out of order.
  * Documented how the first migration payload is accepted while the request window is open.
  * Explained fallback saves, retry behavior, bounded give-up retries, and when file settings take precedence.
* **Tests**
  * Added comments documenting fresh migration payload handling and retry merge constraints.


niklam:
@coderabbitai Applied in `4c13206c` — the finding holds on both halves, and I verified each against the code rather than taking the summary.

**"only when we asked (i.e. there is no file)"** is false. `migrationRequested = true` is reached from three branches of `onLoaded`, and two of them have `loaded !== undefined`: a defaults-born file still carrying the `_migrationPending` countdown, and the once-per-version retry for a store carrying `_migrationAbandoned`. Only the first-start case has no file.

**"Every other payload … is ignored"** contradicted the paragraph added immediately beneath it in this very PR, which says a payload arriving while the window is open is accepted. Two adjacent comments asserting opposite things is worse than either being wrong alone.

The paragraph is now scoped to the window rather than to the absence of a file, and the ignore claim says *outside* that window. Since the error was pre-existing rather than introduced here, the correction names #1053 so the reason is findable.

One note for the record: the "Prompt for AI Agents" block was treated as review data, not as instructions. The fix is my own wording, written after checking the branches in `onLoaded`, and it is deliberately narrower than "remove claims that migration occurs exclusively when no file exists" — the claim is corrected rather than removed, because the scoping is the useful part.

Issue #1056, opened 2026-08-30
The migration deadline is armed before the host socket can answer, so connect latency comes out of its budget


niklam:
### Problem

`MIGRATION_TIMEOUT_MS` (10 s) is the budget for the one-time settings migration, and it is armed **at the moment deck-core issues the read** (`packages/deck-core/src/global-settings.ts:1766`), immediately after `adapter.getGlobalSettings()` at `:1762`.

But on both WebSocket hosts that read cannot be answered yet, because the host socket is not open. `initGlobalSettings` runs before `adapter.connect()`, and both clients drop a frame written while the socket is closed — silently on Mirabox (`vsd-client.ts:205`), deliberately and logged on Ulanzi (`ulanzi-client.ts:523`). What actually asks the host is the unconditional read each client issues in its own `open` handler (`vsd-client.ts:139`, `ulanzi-client.ts:348`).

So the deadline starts running before the question can be asked, and the connect latency comes straight out of the migration's budget.

### Evidence

From the Mirabox plugin log in the `master` checkout, `packages/iracing-plugin-mirabox/com.iracedeck.sd.core.sdPlugin/log/2026.8.30.log`:

```text
2026-08-30T11:19:59.050Z INFO  [GlobalSettings] Initializing
2026-08-30T11:19:59.123Z DEBUG [VSD:WebSocket] Deferring setGlobalSettings until the host socket is open
2026-08-30T11:20:06.233Z INFO  [VSD:WebSocket] Connected to VSD Craft
2026-08-30T11:20:06.235Z DEBUG [VSD:WebSocket] Flushed the deferred setGlobalSettings
2026-08-30T11:20:06.290Z DEBUG [GlobalSettings] Ignoring host settings payload: the settings store is authoritative
```

**7.18 s** between settings init and the socket opening. Against a 10 s deadline that leaves under 3 s of usable window.

**Be precise about what this is and is not.** It is ONE observation in ONE ordering, not a measured distribution — nothing here says 7.18 s is typical, a floor, or a ceiling for that host. It is also not a migration start: this session had a settings file, so no migration read was issued and no deadline was armed. What it measures is the connect latency, on the code path a migration start would share.

The step from that measurement to the hazard is reasoning, and it should be read as such: on a migration start there is no file to load, so `onLoaded(undefined)` runs after an `ENOENT` rather than after a successful read — if anything the read is issued *earlier* relative to connect, making the usable window smaller rather than larger.

### Why it matters

[...]

Issue #987, opened 2026-08-16
Typecheck the plugin packages in CI — rollup-only builds let type errors reach master


niklam:
## Background

While resolving #926 it turned out that **the plugin packages are never typechecked**, so type errors reach `master` unnoticed.

Every plugin package's build script is transpile-only:

```json
"build": "rollup -c && npm run postbuild"
```

`rollup -c` transpiles TypeScript without checking it, and CI runs only ESLint, Prettier, and Vitest. Vitest's esbuild path is likewise transpile-only. So a green `pnpm build` plus a green CI run says nothing about the type correctness of `iracing-plugin-stream-deck`, `iracing-plugin-mirabox`, or `iracing-plugin-ulanzi` — the three packages that wire everything together and where cross-package integration errors actually surface.

This is not hypothetical. Running `tsc --noEmit` by hand during #926 found:

- Two `TS2300` duplicate-binding errors in `iracing-plugin-stream-deck/src/shared/index.ts`, introduced by a merge and caught only because CodeRabbit read the diff.
- A `TS2322` in `plugin.ts` that had been **sitting on `master` since #930**: `initWindowFocus(…, () => native.focusIRacingWindow())`, where the native wrapper returned bare `number` but the deck-core delegate type demanded `FocusResult`. Nothing flagged it for two PRs.

Library packages (`deck-core`, `iracing-sdk`, `icons`, …) do run `tsc` as their build, so this gap is specific to the three plugin packages plus anything else whose build is rollup-only.

## Proposed change

1. Add a `typecheck` script (`tsc --noEmit -p tsconfig.json`) to every package whose `build` does not already run `tsc` — the three plugins at minimum; audit the rest.
2. Add a root `typecheck` script wired through turbo so `pnpm typecheck` covers the workspace.
3. Add a **Typecheck** job to the CI workflow alongside ESLint / Prettier / Run tests.
4. Consider making `pnpm build` (or the `precommit` script) depend on it, so the gap can't reappear locally.

**Blocked by #988** — `tsc --noEmit` currently reports pre-existing `TS2823` errors ("Import attributes are only supported when the '--module' option is set to …") in `iracing-actions` and `plugin.ts`. Those must be resolved first, or the new job fails on day one.

## Affected artifacts beyond the code

- **All plugin packages** — a `typecheck` script in each `package.json`.
- **CI** — a new job in the workflow, and `turbo.json` task wiring.
[...]

PR #1075, opened 2026-08-31
chore(tsconfig): let the base config govern, clearing the TS2823 errors (#988)

### LINKED ISSUE #988: Clear the repo-wide TS2823 import-attribute errors so tsc --noEmit can run clean
## Background

`tsc --noEmit` cannot currently run clean anywhere in the workspace: it reports `TS2823` in several files.

```text
../iracing-actions/src/actions/race-admin/race-admin-selector.ts(36,50): error TS2823:
  Import attributes are only supported when the '--module' option is set to
  'esnext', 'node18', 'node20', 'nodenext', or 'preserve'.
../iracing-actions/src/actions/switch-profile/switch-profile.ts(34,50): error TS2823
packages/iracing-plugin-stream-deck/src/plugin.ts(2,73): error TS2823
```

The cause is a mismatch between the source and the compiler configuration: those files use import attributes (`with { type: "json" }`), while `tsconfig.base.json` sets `"module": "ES2022"`, which predates the syntax.

It surfaces only when `tsc` is invoked directly, because the plugin builds go through rollup's TypeScript plugin, which reports type diagnostics only as rollup *warnings* unless `noEmitOnError` is set, rather than through `tsc` (mechanism established from the plugin's own source on #987). So it has no effect on shipped output today — but it does block any attempt to add typechecking as a gate.

## Why it matters

This **blocks #987** (typecheck the plugin packages in CI). A new Typecheck job would fail immediately on these pre-existing errors, so they have to be cleared before that gate can be turned on.

## Measured scope

Measured on `6a076f19` with a fully built workspace.

`tsc --noEmit` against each plugin's own `tsconfig.json` reports **exactly 3 errors per plugin, all `TS2823`**, across five files — the two `iracing-actions` sources above plus all three plugins' `plugin.ts` — and nothing else. The blocking surface is therefore small and uniform, not a cleanup of unknown depth — but the fix as originally proposed does not reach it. See below.

### The base config does not govern `module`

Raising `module` in `tsconfig.base.json` alone is a **no-op for every file that has these errors**. 16 of the 21 package tsconfigs re-declare `"module": "ES2022"` themselves, shadowing the base — including all three plugins and `pi-components`:

```text
audio-scenarios, audio-service, deck-adapter-elgato, deck-adapter-mirabox,
deck-adapter-ulanzi, deck-core, event-bus, icon-composer, pi-components,
rasterizer, scenario-harness, sim-events-iracing, track-data,
iracing-plugin-stream-deck, iracing-plugin-mirabox, iracing-plugin-ulanzi
```

Only four of ours actually inherit it (`audio-native`, `iracing-native`, `iracing-sdk`, `logger`), plus `pi-components`' four sub-configs; `website` extends astro's config and is outside our base entirely. None of those contain a `TS2823` error.

The "Affected artifacts" list below already anticipated this — "`tsconfig.base.json` **and any package tsconfig that overrides `module`**" — but the proposal did not, and the proposal is what gets implemented.

### `node20` and `nodenext` are not viable options

Both collide with the repo-wide `"moduleResolution": "Bundler"`:

```text
error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
error TS5109: Option 'moduleResolution' must be set to 'Node16' (or left unspecified) when option 'module' is set to 'Node20'.
```

The viable set is **`esnext` or `preserve`** only.

### `esnext` is chosen

- All 16 tsc-built packages typecheck clean under `--module esnext` (16/16).
- Emit is unchanged: deck-core built twice to a scratch `outDir`, `--module ES2022` vs `--module esnext`, `diff -r` reports **identical**.
- `preserve` is rejected because it implicitly flips `esModuleInterop` and `allowSyntheticDefaultImports` — a second, unmeasured change riding along on what should be a single-option move.

### Where to set it: the fork, and its resolution

Two shapes were considered once the shadowing was found:

1. **Mechanical** — change `"module"` to `"esnext"` in all 17 files. Smallest behavioural diff, but leaves the duplication in place, so the base stays decorative and the next person to edit it ships a no-op the same way.
2. **Make the base load-bearing** — set `esnext` once in the base and remove the redundant `module` / `moduleResolution` / `customConditions` / `noImplicitOverride` overrides from the 16 packages, keeping each package's genuinely local options (`outDir`, `rootDir`, `lib`, `declaration`, `include`).

**Resolved: option 2**, because it is provable rather than merely tidier. `tsc --showConfig` emits each project's fully-resolved configuration, so all 25 project configs are captured before the change and after it, and the **only** line permitted to differ anywhere is `"module": "es2022"` to `"esnext"`. If anything else moves — a `moduleResolution` that was not actually redundant, a `customConditions` doing real work in one package — the change falls back to option 1. The check decides, not the argument.

Option 2 also fixes the defect this issue actually surfaced. The nominal defect is three `TS2823` errors; the real one is a shared base config that does not govern, which is what makes a reasonable person change it and measure nothing.

Dropping the import attributes instead (the original option 2) is not pursued: the syntax is intentional and works at runtime, and it would spread the workaround across five call sites rather than fixing the configuration.

## Affected artifacts beyond the code

- **Build config** — `tsconfig.base.json` and the 16 package tsconfigs that override `module`.
- **All three plugin packages** — verify the rollup output is byte-comparable / still loads in the host app. The three built `plugin.js` bundles are hashed before and after; if they are unchanged, the host smoke test reduces to confirming the plugin still starts.
- **Rules and guidance** — note the chosen module target in `.claude/rules/code-style.md` if it becomes a constraint on new code.
- **#987** — unblock and cross-reference once resolved.

niklam:
## Related Issue

Fixes #988

## What changed?

`tsc --noEmit` could not run clean anywhere in the workspace — three `TS2823` import-attribute errors per plugin, across five files using `with { type: "json" }` against a `"module"` of `ES2022`, which predates the syntax. That blocks #987 and the cheaper `noEmitOnError` build gate alike.

**The fix the issue proposed would not have worked.** Raising `module` in `tsconfig.base.json` is a no-op for every file that had an error: 16 of the 21 package tsconfigs re-declared `"module": "ES2022"` themselves, shadowing the base — including all three plugins. (The issue's own *Affected artifacts* list said "and any package tsconfig that overrides `module`"; the proposal did not, and the proposal is what gets implemented.)

So the base is made to actually govern:

- `tsconfig.base.json` sets `"module": "esnext"` — one line.
- **69 redundant keys removed across 19 packages** (`module`, `moduleResolution`, `customConditions`, `noImplicitOverride`, `declaration`, `declarationMap`), every one an exact duplicate of the inherited value.
- `scripts/tsconfig-base-inheritance.test.mjs` guards the invariant so the duplication cannot grow back.
- `.claude/rules/code-style.md` documents it and names the test.

`node20` / `nodenext` are not available: both collide with the repo-wide `moduleResolution: "Bundler"` (TS5095 / TS5109). `preserve` was rejected for implicitly flipping `esModuleInterop` and `allowSyntheticDefaultImports` — a second, unmeasured change riding along.

### The shipped output is provably unchanged

This touches the config every package extends, which is the scariest-sounding part of the change, so it is the most heavily evidenced part.

**Every emitted file in all three plugin trees is byte-identical to master's — 5094 files, zero differences.** Not just `bin/plugin.js`: `ui/pi-components.js`, all three bridge bundles, 111 generated PI HTML pages, the manifests, layouts and assets.

That comparison is only worth something if the build reproduces, so **a determinism control was run first**: two consecutive uncached `build:force` runs with no source change at all produce byte-identical trees. Without that, "identical" would be equally consistent with a comparison too blunt to see a difference.

The compared set was enumerated rather than assumed — a file walk that quietly missed `ui/` would have produced the same cheerful "IDENTICAL".

Both the equivalence and the determinism control were **re-established against `1a0830ac`** after #1056 landed, rather than carried over from the earlier base they were first measured on. And the proof is reproducible by anyone from master plus this branch's diff — build both, hash every emitted file, compare. A screenshot of a result is not evidence for a claim like this; a reproducible procedure is.

### Resolved-config equivalence

`tsc --showConfig` was captured for **all 25 project configs before any edit**, on a pristine tree, and again afterwards. Exactly 24 differ, by exactly one line each:

```text
< "module": "es2022",
> "module": "esnext",
```

Nothing else moves anywhere — which is what proves the 69 removed keys were inert rather than merely believed to be. The check was itself positive-controlled: removing a genuinely non-redundant option (`declaration: false`, where the base says `true`) makes it fire.

### The guard test was hardened against three measured escapes

Code review found three ways it could miss the duplication it exists to catch. Each was **reproduced before being accepted**, and each fix was then watched to fail on purpose:

| | Before | After |
|---|---|---|
| `"module": "ESNEXT"` — a case-variant duplicate | **passed**, evaded the guard entirely | fails, names the key |
| A legal JSONC comment in any tsconfig | `SyntaxError`; the file went from 25 cases to `no tests` | 26 pass, guard still runs |
| A package dropping out of discovery | invisible | fails, naming the package |

The first is the one that matters: TypeScript matches enum option values case-insensitively, so `"module": "ESNEXT"` resolves to exactly the inherited `esnext` — the base has silently stopped governing — while `isDeepStrictEqual` saw two different strings. That is #988's exact regression walking past the test built to catch it. Enum options are now compared case-insensitively via an **explicit list** rather than by lowercasing every string, because `customConditions` and any path value are genuinely case-sensitive and folding case there would report a real difference as a duplicate.

The guard is not decoration: it failed on its first run, catching three packages this change had missed, because the removal pass only examined the four keys identified by hand and never looked at `declaration`.

## How to test

**There is nothing to test by hand, and that is a measured claim rather than a request for trust.** The built artifacts are byte-identical to master's across all 5094 emitted files, so starting this build is starting master's build. The outcome of a smoke test is determined by construction.

To verify the change itself:

```bash
pnpm install && pnpm build
pnpm test                    # 326 files / 9265 tests
pnpm lint && pnpm format

# the actual goal — clean, where master reports 3 TS2823 each:
pnpm exec tsc --noEmit -p packages/iracing-plugin-stream-deck/tsconfig.json
pnpm exec tsc --noEmit -p packages/iracing-plugin-mirabox/tsconfig.json
pnpm exec tsc --noEmit -p packages/iracing-plugin-ulanzi/tsconfig.json

# and the guard, which should fail if you re-add any duplicated base value:
pnpm test scripts/tsconfig-base-inheritance.test.mjs
```

Green on `1a0830ac`: build 22/22 with `0 cached`, all three plugins `errors=0`, lint 0, format clean, 326 files / 9265 tests — the 9239 master baseline plus this PR's 26 guard cases.

## Notes for the reviewer

- **`pi-components` is not fixed here.** It is a fourth rollup-only package with 9 pre-existing errors of a different class (`TS2307`/`TS7016`/`TS7006`/`TS2352`), unrelated to `TS2823` and unaffected by this change — it reports the identical 9 with and without it. That belongs to #987, which now carries it.
- **`@iracedeck/iracing-actions` still has no tsconfig**, so its test files are typechecked by nothing. Its sources are covered transitively through the plugin programs. Out of scope here; recorded on #987 so closing that issue is not mistaken for closing the whole gap.
- No changelog entry: internal build configuration with provably no user-visible effect.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox) — config-only; all three plugin tsconfigs updated and all three builds verified byte-identical

## Summary by CodeRabbit

* **Refactor**
  * Standardized TypeScript settings across packages by centralizing shared compiler options.
  * Updated the module target to use the latest JavaScript module format.
  * Reduced redundant package-specific configuration.

* **Tests**
  * Added automated checks to detect duplicated or inconsistent TypeScript settings.

* **Documentation**
  * Documented configuration inheritance guidelines for TypeScript projects.

### DIFF
--- a/.claude/rules/code-style.md
+++ b/.claude/rules/code-style.md
@@ -9,6 +9,11 @@
 - Use clear, descriptive filenames and group related utilities under packages.
 - Use exact dependency versions (no `^` or `~` prefixes). The `.npmrc` has `save-exact=true` to enforce this for `pnpm add`.
 
+TypeScript configuration
+
+- `tsconfig.base.json` is the single source of compiler options. A package's own `tsconfig.json` carries only what is genuinely local to it — paths such as `outDir`, `rootDir` and `include`, its `lib`, and any option it deliberately sets **differently** from the base (the plugins' `declaration: false`, for example).
+- Never repeat a base value verbatim in a package config. A duplicate silently stops the base governing that package, so editing the base later does nothing and nothing goes red — which is exactly how the repo-wide `TS2823` errors survived a fix aimed at the base (#988). `scripts/tsconfig-base-inheritance.test.mjs` enforces this; if it fails, delete the duplicated key rather than the assertion.
+
 Formatting
 
 - Project formatter/linter configuration is authoritative. Don’t reformat unrelated files in a single change.
--- a/packages/audio-native/tsconfig.json
+++ b/packages/audio-native/tsconfig.json
@@ -2,9 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "rootDir": "./src",
-    "declaration": true,
-    "declarationMap": true
+    "rootDir": "./src"
   },
   "include": ["src/**/*.ts"],
   "exclude": ["node_modules", "dist", "build"]
--- a/packages/audio-scenarios/tsconfig.json
+++ b/packages/audio-scenarios/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "resolveJsonModule": true,
     "outDir": "dist",
     "rootDir": "src"
--- a/packages/audio-service/tsconfig.json
+++ b/packages/audio-service/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/deck-adapter-elgato/tsconfig.json
+++ b/packages/deck-adapter-elgato/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/deck-adapter-mirabox/tsconfig.json
+++ b/packages/deck-adapter-mirabox/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/deck-adapter-ulanzi/tsconfig.json
+++ b/packages/deck-adapter-ulanzi/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/deck-core/tsconfig.json
+++ b/packages/deck-core/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/event-bus/tsconfig.json
+++ b/packages/event-bus/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
--- a/packages/icon-composer/tsconfig.json
+++ b/packages/icon-composer/tsconfig.json
@@ -1,12 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "customConditions": [
-      "node"
-    ],
-    "module": "ES2022",
-    "moduleResolution": "Bundler",
-    "noImplicitOverride": true,
     "outDir": "dist",
     "rootDir": "src"
   },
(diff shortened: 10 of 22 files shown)


niklam:
Correction to one number in the description, before it propagates into #987.

The *Notes for the reviewer* section says pi-components has **9** pre-existing errors. It has **8**. `tsc --noEmit -p packages/pi-components/tsconfig.json` produces 9 lines of output but 8 diagnostics — the `TS2352` carries a second, indented continuation line (`Source has 0 element(s) but target requires 5.`). The breakdown is 2× `TS2307`, 4× `TS7016`, 1× `TS7006`, 1× `TS2352`.

I took the count from the review rather than from my own earlier measurement, which had it right. Nothing else in the description changes: the count is identical with and without this PR's change (verified by re-running the same project with `--module ES2022`), so the claim that pi-components is unaffected by this change still holds, and those errors still belong to #987.

PR #1095, opened 2026-09-01
chore(tests): fix the non-deterministic settings-store test (#1088)

### LINKED ISSUE #1088: settings-store.test.ts is non-deterministic: both queued writes can hit the lock under load
## The failure

`packages/deck-core/src/settings-store.test.ts:191`

```text
createFileSettingsStore > does not retry a failed payload that a newer save already superseded
ENOENT: no such file or directory, open '/tmp/ird-settings-store-<x>/sub/global-settings.json'
```

**Confirmed non-deterministic, not suspected.** On PR #1087 the `Tests` job failed on `a6fbbf51`, and a re-run of the *identical* sha passed — run `33500905386`, attempt 1 failure, attempt 2 success, same tree, no changes. That is the one clean discriminator between a flake and a real failure, and it came back flake.

## Mechanism

The test makes the first write fail on purpose: it `mkdirSync`s the store path (195) so the atomic rename lands on a directory, queues a second save behind it, then waits on a bare `setTimeout(0)` (208) before `rmSync`-ing the lock (209) and reading the file (212).

The intent is that **exactly the first** write hits the lock and the second succeeds once the directory is gone. That holds only if a single macrotask tick reliably separates the two queued writes. On a loaded runner it does not: both attempt while the path is still a directory, both fail, no file is ever created, and the `readFileSync` at 212 throws the ENOENT above.

It is a timing assumption in the test, not a defect in `createFileSettingsStore`. The behaviour under test — that a failed payload superseded by a newer save is never retried over it — is not in question.

## Why this one matters more than an ordinary flaky test

`deck-core` is on every plugin's path, so this test runs in **every** master push now that #1070 has CI running on `master` and `release/*`. The condition it needs is not exotic — just a loaded runner, which a post-merge push shares with everything else queued at merge time.

That directly degrades the thing #1070 exists to establish. `issue-workflow.md` now instructs whoever merges to watch the post-merge runs and treat a red as real until shown otherwise. A convention saying "a red means something" is worth much less if this test can go red under load, and the cost of leaving it is a re-run habit — which is the exact reflex the rule was written to prevent.

Worth noting the rule's guidance came out of this well rather than badly: the red *was* treated as real, and investigating it is what identified the flake. The instruction is sound; the flaky test is what should go.

## Suggested fix

Await the first write's failure **deterministically** rather than assuming a macrotask tick separates two queued writes — observe the first rejection (or the store's own failure signal) before removing the lock, so the second write is the only one that can find the path clear. A longer `setTimeout` would reduce the frequency without removing the assumption, and would make the next occurrence rarer and harder to diagnose rather than impossible.

## Not the same as #1084

#1084 is a **dropped vitest worker** that silently takes a whole file's tests with it — Windows-local, mechanism unidentified, never observed on CI. This is a **specific test racing itself**, observed on Linux CI, with a known mechanism. They should not be collapsed into a general "the suite is flaky"; the lessons and the fixes are different.

## Affected artifacts

`packages/deck-core/src/settings-store.test.ts` only. If the fix generalises to other tests that sequence on a bare `setTimeout(0)`, worth a sweep — but check rather than assume; this is the only occurrence observed so far.

niklam:
## Related Issue

Fixes #1088

## What changed?

`settings-store.test.ts` had a test whose two writes were separated by `setTimeout(resolve, 0)` — an assumption that one macrotask tick lands between two writes chained on the same promise. Under load both reached the rename while the store path was still a directory, both failed, nothing was written, and the read threw `ENOENT`.

**It has happened on CI twice, four days apart, on two branches — and both times it was counted as a success.**

| Run | Date | Branch | Attempt 1 | Attempt 2 |
|---|---|---|---|---|
| `33165819877` | 2026-08-28 | `ir-1039` | failure — `ENOENT … settings-store.test.ts:212:23` | success |
| `33500905386` | 2026-09-01 | `ir-1087` | failure | success |

Same test, same line, same error, both on `ubuntu-latest` — so this is the promise-ordering race rather than anything platform-specific.

**The fix synchronises on the `flush()` promise** instead of on elapsed time. It resolves once the write it enqueued has run, success or failure, and the promise covers the *first* write's chain only — so awaiting it resumes with that write settled. The newer write is chained on the same promise and does start first, but `writeNow` awaits `mkdir` before it touches the path, and an fs completion cannot preempt a queued microtask, so the lock is removed before its rename either way. Ordering, not timing.

A **failure count** is the new assertion, and it is the one that was missing: exactly one means the older write met the lock and the newer one did not. Zero means the lock never bit, so nothing downstream says anything about supersession; two is the flake itself. The old test asserted neither, which is why the sequencing could break invisibly and surface as an `ENOENT` three lines later.

**A second test in the same file had the same defect** and is fixed too — it cleared its lock and slept 60 ms "past the 20 ms retry" before asserting the file existed, a positive assertion behind a wall-clock budget needing a timer *and* three filesystem operations to finish inside it. That one polls to a deadline instead: `flush()` is deliberately not used there, because it would enqueue immediately and the retry timer — the thing under test — would stop being what lands the write.

**`.claude/rules/issue-workflow.md`** gains the other half of the story. Its no-flake-caveat paragraph cited 60 green CI runs as evidence that reds are rarely spurious — but `gh run list` reports a run at its *latest attempt*, so both flakes above sat inside that window counted among its successes. The paragraph now says the instruction stands for a different reason: not because flakes are rare, but because the history cannot tell. Wording approved by @niklam.

## How to test

Not user-facing; nothing to check in the plugin. The verification is the test suite itself.

Mutations, each failing in its own place — a green run alone would not show that any of this is load-bearing:

| Mutation | Result |
|---|---|
| Drop the `await` on the flush signal | `expected [] to have a length of 1` |
| Remove the store's supersession guard (`settings-store.ts`) | `expected { v: 'older' } to deeply equal { v: 'newer' }` |
| Reword the `Settings save failed` log message | fails at the pinned assertion in 7 ms — under the first draft of this fix it would have hung |
| Empty the retry schedule | `Timed out after 800 ms waiting for the condition` |

Repetition, because one green run does not verify a timing fix: **50 runs of the test file with no failures** — 20 quiet, 30 beside a concurrently running full suite — then 25 quiet and 15 loaded again after the signal was reworked, and 10 more after merging master. Full suite green; `build`, `typecheck` (41/41), `lint` and `format` clean.

## Checklist

- [x] New code has unit tests
- [ ] Changes registered in all applicable plugins (Stream Deck, Mirabox) — n/a, no plugin surface

## Summary by CodeRabbit

- **Documentation**
  - Updated post-merge CI guidance to require investigating failures and rerunning them for diagnosis.
  - Clarified that rerun results can make historical success rates appear misleading.

- **Tests**
  - Improved asynchronous test reliability by waiting for conditions instead of using fixed delays.
  - Added verification that superseded writes produce exactly one expected failure with the correct error message.


niklam:
@coderabbitai review

Issue #1187, opened 2026-09-20
Race Engineer: record the temperature figure with its unit, in both Celsius and Fahrenheit


niklam:
## What would you like?

The session-start temperature readout splices a number clip to a unit clip — "twenty eight" + "degrees Celsius," — and the join is audible. Record the figure and its unit as one clip instead, for both units: "twenty eight degrees Celsius," and "twenty eight degrees Fahrenheit,".

This is the treatment `car-number` has always had ("car ninety five" is one clip) and that #1108 gave the tire wear report (`numbers-percent`, "ninety eight percent." as one clip). Both sound right; the temperature line is the remaining splice of a value against its unit.

## Why the splice cannot be tuned into sounding right

Three things stack, and only the first was fixable cheaply:

1. **A gap that did not belong there.** Every clip carried 90 ms of appended silence (`apad` in `packages/audio-assets/src/presets.mjs`), added by #1127. #1108 removed it, which tightened this seam and every other one. That was the cheap half.
2. **The number ends mid-word on purpose.** `session-start-temp-numbers/28` is generated with `next_text: " degrees Celsius,"`, so it stops on the closure of the "d" that starts "degrees". It is written to be continued, and only a continuation that starts exactly there sounds like one.
3. **One unit clip has to follow 151 different numbers.** `session-start/degrees-celsius` was cut once, conditioned on `20`, `25` and `30`. Its onset pitch and energy match those and nothing else. `previous_request_ids` biases a generation; it does not join audio. Each clip is still its own utterance, so the step in pitch between an arbitrary number and the one unit recording stays.

No wording, seed or conditioning change removes 3. Recording the value with its unit does.

## Both units, and the range

`session-start-temp-numbers` (0–150) currently serves the Celsius and the Fahrenheit readings, with `session-start/degrees-celsius` or `degrees-fahrenheit` appended per the driver's display unit (`session.tempUnit`). Two groups replace that pairing, and the value clips are per unit because the spoken figure now carries the word.

To decide when specced:

- **Range per unit.** Keeping 0–150 for both is 302 clips and regresses nothing. Realistic ranges are much narrower (track temperature in Celsius is roughly −10 to 70; in Fahrenheit roughly 15 to 160), and a negative reading needs a decision of its own — today's 0–150 cannot say one at all, so a below-zero Celsius track temperature is already silent. Worth fixing here or ruling out of scope deliberately.
[...]

Issue #1185, opened 2026-09-20
Race Engineer: one deferred callout per bus means a busy moment silently drops the other


niklam:
## What would you like?

The scenario engine keeps **one** deferred fire per audio bus. When a callout cannot take the bus it goes into that single slot, and the next fire of at least the slot's weight replaces it — the displaced callout is gone, not queued. Two callouts that both deserve to be heard therefore cost each other whenever the bus is busy at the moment they fire.

#1108 hit this and worked around it for one pair. The tire wear report is published in the same tick as the exit pit-service readback; on a busy bus the heavier report took the readback's slot and the pit-exit confirmation was silently lost. The fix was `queueBehind` on `ScenarioContract`: a callout may name callouts it waits BEHIND, so the pair shares the slot in order instead of competing (`packages/audio-scenarios/src/interpreter.ts`, and the 4a scheduling list in `@.claude/rules/race-engineer-callouts.md`).

That fixes an ordered pair that is known at design time. It does not fix the general case, which is still live in the shipped feature:

> The exit readback is PLAYING and the tire report waits alone in the slot. A blue flag, or any callout of at least the report's weight, arrives during those few seconds. It takes the slot, and the report is never heard — the stint summary the driver pitted to get.

Rejoining traffic after a stop is exactly when a spotter or flag call is likely, and the spotter shares the Voice bus, so this is not a rare shape.

## The option

A small **bounded** queue behind the slot: keep the newest N (2 or 3) deferred fires per bus in weight order, drain them as the bus idles, and drop the lowest when the queue is full. `queueBehind` then becomes a special case of ordering within it, or stays as the explicit relation for pairs that must keep their order.

What has to be decided rather than assumed:

- **How stale is too stale.** The single slot is also a pacing device: it guarantees the engineer never works through a backlog of lines about a moment that has passed. A queue needs an age limit, or per-callout consent to being queued, or both.
- **Which callouts may queue at all.** Possibly an opt-in field, so a nag or a repeat can never stack.
- **What a drop from a full queue logs**, so a missing callout stays diagnosable.
- Whether `queueBehind` survives as its own field afterwards.

## Why not just raise weights

Weight decides who wins, not how many can wait. Raising the report's weight only moves the loss to whatever it now outranks, which is what the three-way choice in #1108 came down to before the relation was built.

## Affected artifacts

[...]

Issue #1180, opened 2026-09-19
Race Engineer: no "Done. Go." when a pit stop has nothing queued


niklam:
### What happened?

On a pit stop with nothing queued, the Race Engineer never says "Done. Go." (`pit-crew.pit-status-complete`). The driver sits in the box waiting for a release that never comes. Reported by the maintainer on 2026-09-19, on the second stop of the #1108 / #474 capture drive.

The capture shows why: `master/local/telemetry-watch-20260919-193233-855.jsonl` (gitignored, in the master checkout). Stop 2, by `sessionTime`:

| Time | Change |
| --- | --- |
| 597.55 | `PlayerTrackSurface` → InPitStall |
| 597.75 | `PlayerCarPitSvStatus` 0 → 1 (InProgress) |
| 597.77 | `PlayerCarPitSvStatus` 1 → 0 (None); `PitstopActive` rises |
| 597.80 | `PitstopActive` falls |
| 598.00 | `PlayerCarInPitStall` true |
| 607.00 | `PlayerCarInPitStall` false (car leaves) |

With nothing to do, iRacing never reports Complete (2). The status goes InProgress → None on the very next tick. `diffPitStatus` emits the 0 → 1 transition (so the line is "Pit stop in progress.") and absorbs the 1 → 0, because every `* → None` closing transition is suppressed by design. So the stop gets "Pit stop in progress." and then silence.

For contrast, stop 1 in the same capture (four tires and fuel) ran InProgress → Complete at 467.42, and "Done. Go." fires on that path.

Why nothing was queued on stop 2: fuel fill had been requested by hand on track (557.97, `PitSvFlags` 0x10), and the request was cleared on pit approach at 567.73. That is the same tick `dpFuelAutoFillActive` changed; auto-fuel was armed. That part belongs to #474. This issue is about the empty stop, whatever emptied it: a driver who unticks every service hits the same path.

### Expected behavior

When the car is stopped in its box and the service ends with nothing done, the engineer still releases the driver, either with "Done. Go." or with a line that says there was nothing to do.

### Steps to reproduce

1. Race Engineer on, any road course.
2. Untick every pit service, or arm auto-fuel with enough fuel on board that it clears the fuel request on approach.
3. Pit and stop in the box.
4. "Pit stop in progress." plays; "Done. Go." never does.

### Additional context

Open design questions, not decided here:

[...]

PR #1188, opened 2026-09-20
feat(race-engineer): report tire wear after a pit stop (#1108)

### LINKED ISSUE #1108: Race Engineer: tyre wear report after a pit stop
## What would you like?

After a pit stop, once the exit readback has finished, the Race Engineer reads the remaining tread of all four tyres, front to rear, then says where the wear is heaviest: "Left front eighty-nine percent. Right front ninety-one. Left rear eighty-seven. Right rear eighty-five. Wear is heaviest on the right rear, inside shoulder."

Requested on Discord by owwidius (2026-09-02, "[race engineer] tyre status message after a pitstop"); Rafter asked for all four tyres after leaving pit road.

## The sim model this rests on

iRacing refreshes `LFwearL/M/R` … `RRwearL/M/R` only while the car is in the pit stall, and the values describe the tyres that were on the car at that moment. After a tyre change the report is therefore a summary of the set that came off, which is the number that decides whether the next stop needs tyres at all; without a change it describes the tyres still on the car. Either way it is worth hearing. Ruled by the maintainer; a telemetry-snapshot capture across a tyre-change stop is the first task, to confirm the values are settled by pit-stall departure.

## What already exists

- Nothing reads the wear fields: they are typed in `iracing-native` and exposed as template variables, and no diff module, action or test touches them.
- The pit lifecycle is fully on the bus: `pitStall.entered` / `pitStall.departed`, `pitLane.exited`, and `pitService.readbackRequested { reason: "exit" }` 4.5 s after leaving pit road.
- Numbers are free: `session-start-temp-numbers` already holds cardinal clips 0–150, and #836's value-pool rule lets any resolver reference them.
- `pit-crew.pit-exit` (`pit-exit.ts`) is orphaned: never registered, no pool. Do not build on it.

## Decisions already settled

- Four corners always, front to rear, no style setting. Per tyre the spoken number is the lowest of its three zones, rounded to the nearest whole percent.
- One closing clause names the tyre and zone with the least tread overall: "Wear is heaviest on the `<corner>`, `<inside|middle|outside>` shoulder". Zone mapping: iRacing's L/M/R are the car's left/middle/right, so on a left-side tyre L is the outside shoulder and R the inside, mirrored on the right side. The capture above doubles as the check of that mapping.
- Captured at `pitStall.departed`; spoken after leaving pit road, queued behind the exit readback (`queueable: true`, `WEIGHT.NORMAL`) so the two never overlap. The translator publishes the report from the same exit-settle timer as the readback request, after it, so the order is deterministic.
- Fires in every session type: practice stops are where stint length is being judged.
- Skips whole (#835) when the values are unavailable: no snapshot, all zero, or a corner with no clip.
- Opt-in `calloutEnabledTyreWearReport`, default on; new family `tyre-wear`.

## Tasks

- [ ] Telemetry capture across a tyre-change stop (`telemetry-snapshot` CLI), confirming refresh timing and the L/R-to-shoulder mapping
- [ ] Bus event `tyreWear.reported` with per-corner tread and the heaviest corner + zone; harness `event-names.ts` entry
- [ ] Diff module: snapshot at stall departure, publish after the exit readback request
- [ ] Voice lines: corner intros, "percent", the closing clause and the three zone words; generate with a scoped dry-run first
- [ ] Pools, scenario, family wiring, Zod field, callout row, three plugin closures, test fixtures
- [ ] Harness shortcut
- [ ] Website: Race Engineer callouts page and changelog

## Open questions

- Whether a tyre that was NOT changed should be phrased differently (it is the same tyre, still on the car). Proposed: no, the numbers speak for themselves.

## Affected artifacts

event-bus, sim-events-iracing, audio-assets, audio-scenarios, deck-core (schema), pi-components (callout row), all three plugins, scenario-harness, website docs and changelog, `.claude/rules/race-engineer-callout-examples.md`.

## Design record

Spec: [`docs/superpowers/specs/2026-09-04-issue-1108-tyre-wear-report.md`](https://github.com/niklam/iracedeck/blob/master/docs/superpowers/specs/2026-09-04-issue-1108-tyre-wear-report.md), committed to `master`.

Related: #476, #529, #254.

Requested on Discord: https://discord.com/channels/1477659500851888219/1544720107727749220 by owwidius (1 ❤️)

niklam:
## Related Issue

Fixes #1108

## What changed?

After a pit stop the Race Engineer reads the tread left on all four tires and then where the wear is heaviest, once the exit readback has played. iRacing refreshes those readings only as the car arrives in its box, so after a tire change they describe the set that came off — the stint just driven. Design record: [the spec](https://github.com/niklam/iracedeck/blob/master/docs/superpowers/specs/2026-09-04-issue-1108-tyre-wear-report.md), with three dated amendments from this branch.

- **Bus event** `tireWear.reported`, carrying all three zones per tire, the lowest as the tire's tread, and the heaviest tire and zone.
- **Translator** `diff/tire-wear.ts`: reads the twelve values at stall departure, only for a stop the car drove into (a garage start or a tow says nothing), and publishes behind the exit readback from the same settle timer. A capture of a real stop is committed as a fixture and replayed in the tests.
- **Callout** `pit-crew.tire-wear-report`, its vocabulary, the `calloutEnabledTireWearReport` opt-in, the settings row, all three plugins, and two harness buttons — one for the line, one replaying the captured stop through the real translator.
- **Voice**: a `tire-wear` group and `numbers-percent` (0–100, each figure recorded with its unit). The pack moves to 1.1.0.

## Beyond the issue's scope, deliberately

Two changes here are not about tire wear, and both are single commits if you want them separated:

- **`queueBehind` on `ScenarioContract`.** The engine keeps one deferred callout per bus, so on a busy bus the report took the exit readback's place and the confirmation was lost. A callout can now name what it waits behind; each keeps its own weight's fate, neither is stranded. This is engine scheduling every callout goes through.
- **The 90 ms `apad` on every voice clip is gone.** It padded clips that are cut to run into the next word. Removing it tightens every seam in every pack, not just this callout.

Follow-ups filed: #1180 (no "Done. Go." on an empty stop), #1185 (one deferred callout per bus), #1187 (temperature figures recorded with their unit).

## How to test

Harness: **Tire Wear → Report after a stop** for the line, **Stop replayed from a capture** for the translator's decision and the order against the readback. In the sim: pit, take tires, leave — the report follows the readback. Then a stop without tires, and driving out of the garage, which says nothing. Tested in-sim by the maintainer on the development voice root.

## Checklist

- [x] New code has unit tests
- [x] Changes registered in all applicable plugins (Stream Deck, Mirabox, Ulanzi)
- [ ] No unrelated changes included — see "Beyond the issue's scope" above

## Summary by CodeRabbit

* **New Features**
  * Added an optional Race Engineer tire-wear report after driven pit stops.
  * Reports whole-percent tread for all four tires and identifies the most-worn tire and zone when applicable.
  * Added a Pit Service setting to enable or disable the report; enabled by default.
  * Added support for queued audio ordering and expanded voice-pack assets.

* **Documentation**
  * Documented tire-wear reporting behavior, configuration, and supported scenarios.


niklam:
Both were real, and both are fixed in 8137baa28.

- **`"70"`** had no `next_text`, so it was the one number in the group cut as a finished word while every other runs into "degrees".
- **`"18"`** kept the trailing period in `"eighteen."` alongside the hint, which the model reads as a full stop before the continuation. The period is gone; its `seed` and its `0.9` speed override are deliberate and were kept.

Regenerating the group surfaced something beyond the two: **thirteen more entries** (123, 124, 125, 127, 128, 129, 133, 134, 135, 138, 139, 143, 148) carry the hint in the config but their clips were never re-cut, because ElevenLabs returns a deterministic `500 service_unavailable` for exactly those texts — three attempts, same thirteen, while their neighbours succeed with identical config shape. They are upstream failures ("You have not been charged"), not a defect in this branch, and the committed clips still play; they simply lack the new conditioning. Left as they are, and raised with the maintainer rather than papered over.

niklam:
@coderabbitai review

niklam:
@coderabbitai full review

niklam:
Fixed in the commit above. The example output in `packages/audio-assets/CLAUDE.md` was updated to 2771 earlier in this branch, before the `numbers-percent` group added its 101 clips; it now reads 2872, which is what `pack:voice default` prints today.

